PR #78, opened 2023-02-15
Ensure generated code passes `mypy --strict`

huonw:
This adjusts the code _generated_ by ariadne-codegen to pass `mypy --strict` (https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=strict#strict-mode-and-configuration). 

Having the generated code pass `mypy --strict` allows easier integration into projects that mypy with stricter settings than the default. Without this change, we could not easily use ariadne-codegen, because we have stricter mypy settings, and so mypy was rejecting the generated files, and we need to adding exclusions.

The changes are mostly just adding new annotations that were missing, or adjusting them (especially `dict` -> `dict[str, Any]`). Specifically:

1. annotations in the static files in `ariadne_codegen/generators/dependencies/` 
2. annotations in dynamically generated code:
   1. the `qql` lambda is now as a full `def gql(q: str) -> str: return q` function, so that it can have annotations
   2. `variables: dict = ...` is now `variables: dict[str, object] = ...`
3. fixing the `custom_base_client` test to also pass `mypy --strict`
4. adding `mypy --strict` to run over the generated code in CI, to ensure this keeps up to date

This _is_ a bit annoying, but hopefully it's not too bad, because the generator code (i.e. the `ariadne_codegen` folder, minus the `dependencies`) can continue to be non-strict. The pay-off is much simpler integration!

(Thank you for ariadne-codegen, very cool!)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,5 +27,8 @@ jobs:
       run: |
         pylint ariadne_codegen tests
         mypy ariadne_codegen --ignore-missing-imports
+        # these test outputs represent things that may be committed into other projects that may
+        # have strict mypy settings, and so they should be maximally annotated.
+        mypy --strict tests/main/*/expected_client
         black --check .
         isort . --check-only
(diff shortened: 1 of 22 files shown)


mat-sop (approved):
Thank you! ❤️

huonw:
Thank you for merging so quickly!

PR #79, opened 2023-02-15
Set the full URL in base clients

huonw:
This replaces the `base_url` parameter to `BaseClient` and `AsyncBaseClient` with `url`. This is used as the literal thing to post the request to, to allow customising the URL path. This generalises the previous behaviour of always posting to the `/graphql/` endpoint of the base URL.

For instance, many servers use `/graphql` (without the trailing `/`) and reject requests to `/graphql/`, like GitHub's API (https://docs.github.com/en/graphql/guides/forming-calls-with-graphql#the-graphql-endpoint). This PR allows `Client(url="https://api.github.com/graphql", ...)` to handle this sort of server.

Existing callers like `Client(base_url="https://some_host")` need to be updated to `Client(url="https://some_host/graphql/")`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ from graphql_client.client import Client
 
 Client (with default base client), takes passed headers and attaches them to every sent request.
 ```py
-client = Client("https://example.com", {"Authorization": "Bearer token"})
+client = Client("https://example.com/graphql", {"Authorization": "Bearer token"})
 ```
 
 For more complex scenarios, you can pass your own http client:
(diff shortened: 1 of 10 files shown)


huonw:
Ah, good idea. I see two options for this:

1. have the base client accept both a base URL and an endpoint `def __init__(base_url: str = "", ..., endpoint: str = "/graphql/")`, meaning one calls it like `Client(base_url="https://some.host/", endpoint="/graphql")`
2. have the base client just accept the full URL `def __init__(url: str = "")` and so one calls it like `Client(url="https://some.host/graphql")`

Option 1 isn't a breaking change, but feels a bit more clunky. Option 2 _is_ a breaking change, but seems a bit smoother.

I'll go with option 1 for now, but let me know if you prefer option 2, or something else!

rafalp:
We should go with single full url and break things now that userbase is not too big ;)

huonw:
Done! As I'm sure you can see from the code, I've removed how the base clients set `base_url`, and instead just posted directly to the provided URL.

mat-sop (approved):
Thank you! ❤️

huonw:
Woohoo, thank you for merging!

PR #88, opened 2023-03-02
DOC STRINGS FIXED & METHOD REFACTORED

kuzmich321:
* doc strings corrected
* method refactored to have less amount of code

### DIFF
--- a/ariadne_codegen/generators/client.py
+++ b/ariadne_codegen/generators/client.py
@@ -38,7 +38,7 @@ class ClientGenerator:
         self._data_variable = "data"
 
     def generate(self) -> ast.Module:
-        """Generate module with class definistion of grahql client."""
+        """Generate module with class definition of grahql client."""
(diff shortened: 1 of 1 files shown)


ariadne_codegen/generators/client.py
```diff
@@ -63,23 +63,14 @@ def add_method(
...
-            )
-        )
+        kwargs = {
+            "name": name,
+            "return_type": return_type,
+            "arguments": arguments,
+            "arguments_dict": arguments_dict,
+            "operation_str": operation_str,
+        }
+        method_def = (self._generate_async_method if async_ else self._generate_method)(**kwargs)
```
rafalp, line 73: To be honest this code is even harden to follow than one before change.

kuzmich321: Got it

mat-sop (approved):
Thank you!

PR #90, opened 2023-03-08
Default value for optional arguments

### LINKED ISSUE #86: [feature request] configuration option to assign default None value to optional fields in operations functions
Hey! 

Would it be possible to add a bool config option like `assign_none_to_optional_params` which would determine wether the client will have default `None` value already assigned to operation`s function parameters?

Let's take an example function:

```
    def some_function(
        self,
        amount: float,
        budget_id: Optional[int],
        memo: Optional[str],
        recipient_id: Optional[int],
        payment_account_id: Optional[str],
        recurring: Optional[bool],
        source_id: int
   )
```

`amount` and `source` are mandatory, rest is optional but obviously it will raise an error that those params were not passed to the function.

We'd love to have an option to already assign these to `None` so we just pass the mandatory fields and eventually any other optional ones.

WDYT? Would it be beneficial (and possible) from your point of view to add such option?

mat-sop:
This pr adds default `None` to optional methods arguments. resolves #86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## UNRELEASED
 
 - Fixed generating models from interfaces with inline fragments.
-- Added default `None` values for optional operation arguments.
+- Added default `None` values for generated methods optional arguments.
 
 
 ## 0.3.0 (2023-02-21)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 - Fixed generating models from interfaces with inline fragments.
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 - Fixed generating models from interfaces with inline fragments.
+- Added default `None` values for optional operation arguments.
```
rafalp, line 6:
```suggestion
- Added default `None` values for generated methods optional arguments.
```

rafalp (approved):

PR #97, opened 2023-03-16
Support env variables for remote schema headers values

### LINKED ISSUE #91: Support authorization in schema request from env variable
Hello and thanks for this awesome GraphQL client the Python world was missing so far!

We (aim to) use for our internal GraphQL APIs that are protected by an authorization header, just like in the example,
```
remote_schema_headers = {"Authorization" = "Bearer: token"}
```
which works nicely. The issue with this approach is that the token is in the _pyproject.toml_ file that is known to git. Hence, quite some care is required for inserting the token and removing it again before any changes in the _pyproject.toml_ (e.g., other ariadne-codegen settings) are commited. It would be nice if we could point ariadne-codegen to take the credentials (headers) from elsewhere. As the project is using Pydantic, Pydantics BaseSettings might be a good fit. 

At least in our case, the credentials the generated client needs are the same as ariadne-codegen requires for the introspection query. Perhaps, some kind of _globally_ configured headers that are added to all request could make working with ariadne-codegen even more convenient.

mat-sop:
This pr adds support for env variables for remote schema headers values. Proposed in #91 `remote_schema_headers = {"Authorization" = $ENV_VAR_NAME}` is not a valid time entry. Instead, if the value starts with `$` then is treated as env variable -> `remote_schema_headers = {"Authorization" = "$ENV_VAR_NAME"}`

resolves #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ One of the following 2 parameters is required, in case of providing both of them
 
 Optional settings:
 
-- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`
+- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$BEARER_TOKEN"}`
 - `target_package_name` (defaults to `"graphql_client"`) - name of generated package
 - `target_package_path` (defaults to cwd) - path where to generate package
 - `client_name` (defaults to `"Client"`) - name of generated client class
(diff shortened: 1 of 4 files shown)


rafalp (approved):

README.md
```diff
@@ -46,7 +46,7 @@ One of the following 2 parameters is required, in case of providing both of them
 
 Optional settings:
 
-- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`
+- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$BEARER_TOKEN"}`
```
rafalp, line 49:
```suggestion
- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$AUTH_TOKEN"}`
```

PR #112, opened 2023-03-29
Add remote_schema_verify_ssl option

### LINKED ISSUE #105: Support disabling SSL cert verification for given remote GraphQL schemas
[Discussed here](https://github.com/mirumee/ariadne/discussions/1061)

It should be possible to disable ssh verification in pyproject.toml for given schema in cases where SSL cert is self-signed (eg. running https over internal network).

mat-sop:
resolves #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ One of the following 2 parameters is required, in case of providing both of them
 Optional settings:
 
 - `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer: token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$AUTH_TOKEN"}`
+- `remote_schema_verify_ssl` (defaults to `true`) - a flag that specifies wheter to verify ssl while introspecting remote schema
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -220,7 +221,7 @@ Instead of generating client, you can generate file with a copy of GraphQL schem
 ariadne-codegen graphqlschema
 ```
 
-`graphqlschema` mode reads configuration from the same place as [`client`](#configuration) but uses only `schema_path`, `remote_schema_url`, `remote_schema_headers` options with addition to some extra options specific to it:    
+`graphqlschema` mode reads configuration from the same place as [`client`](#configuration) but uses only `schema_path`, `remote_schema_url`, `remote_schema_headers`, `remote_schema_verify_ssl` options with addition to some extra options specific to it:
```
rafalp, line 224:
```suggestion
`graphqlschema` mode reads configuration from the same place as [`client`](#configuration) but uses only `schema_path`, `remote_schema_url`, `remote_schema_headers` and `remote_schema_verify_ssl` options with addition to some extra options specific to it:    
```

rafalp (approved):

Issue #114, opened 2023-03-30
Top level list variables are not parsed correctly.


strue36:
For example with this query.

```GraphQL
query Abc($v1: [TestModel1!]) {
    abc(v1: $v1){
        field1
    }
}
```

`_convert_dict_to_json_serializable` does not call `dict()` on the elements in the list.

I have a fix I will raise a PR for.

strue36:
There is a similar issue with missing optional top level variables that I will address in the same PR.

PR #116, opened 2023-03-30
Handle top level lists

strue36:
Fixes issue described in #114.

Uses remap from the boltons package (https://boltons.readthedocs.io/en/latest/iterutils.html#boltons.iterutils.remap) to iterate over the dictionary of variables and recursively look into lists to call `dict()` on each model within each list.

Also removes `None`s from the dictionary to handle optional inputs.

### DIFF
--- a/ariadne_codegen/client_generators/dependencies/async_base_client.py
+++ b/ariadne_codegen/client_generators/dependencies/async_base_client.py
@@ -69,12 +69,14 @@ class AsyncBaseClient:
 
         return cast(dict[str, Any], data)
 
+    def _convert_value(self, value: Any) -> Any:
+        if isinstance(value, BaseModel):
+            return value.dict(by_alias=True)
+        if isinstance(value, list):
+            return [self._convert_value(item) for item in value]
+        return value
+
     def _convert_dict_to_json_serializable(
         self, dict_: Dict[str, Any]
     ) -> Dict[str, Any]:
-        return {
-            key: value
-            if not isinstance(value, BaseModel)
-            else value.dict(by_alias=True)
-            for key, value in dict_.items()
-        }
+        return {key: self._convert_value(value) for key, value in dict_.items()}
--- a/ariadne_codegen/client_generators/dependencies/base_client.py
+++ b/ariadne_codegen/client_generators/dependencies/base_client.py
@@ -67,12 +67,14 @@ class BaseClient:
 
         return cast(dict[str, Any], data)
 
+    def _convert_value(self, value: Any) -> Any:
+        if isinstance(value, BaseModel):
+            return value.dict(by_alias=True)
+        if isinstance(value, list):
+            return [self._convert_value(item) for item in value]
+        return value
+
     def _convert_dict_to_json_serializable(
         self, dict_: Dict[str, Any]
     ) -> Dict[str, Any]:
-        return {
-            key: value
-            if not isinstance(value, BaseModel)
-            else value.dict(by_alias=True)
-            for key, value in dict_.items()
-        }
+        return {key: self._convert_value(value) for key, value in dict_.items()}
--- a/tests/client_generators/dependencies/test_async_base_client.py
+++ b/tests/client_generators/dependencies/test_async_base_client.py
@@ -67,6 +67,41 @@ async def test_execute_parses_pydantic_variables_before_sending(mocker):
     }
 
 
+@pytest.mark.asyncio
+async def test_execute_correctly_parses_top_level_list_variables(mocker):
+    class TestModel1(BaseModel):
+        a: int
+
+    fake_client = mocker.AsyncMock()
+    client = AsyncBaseClient(url="base_url", http_client=fake_client)
+    query_str = """
+    query Abc($v1: [[TestModel1!]!]!) {
+        abc(v1: $v1){
+            field1
+        }
+    }
+    """
+
+    await client.execute(
+        query_str,
+        {
+            "v1": [[TestModel1(a=1), TestModel1(a=2)]],
+        },
+    )
+
+    assert fake_client.post.called
+    assert len(fake_client.post.mock_calls) == 1
+    call_kwargs = fake_client.post.mock_calls[0].kwargs
+    assert call_kwargs["url"] == "base_url"
+    assert call_kwargs["json"] == {
+        "query": query_str,
+        "variables": {"v1": [[{"a": 1}, {"a": 2}]]},
+    }
+    assert not any(
+        isinstance(x, BaseModel) for x in call_kwargs["json"]["variables"]["v1"][0]
+    )
+
+
 @pytest.mark.parametrize(
     "status_code, response_content",
     [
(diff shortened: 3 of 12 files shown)


strue36:
> Also removes Nones from the dictionary to handle optional inputs.

This isn't strictly necessary so happy to remove this bit.

ariadne_codegen/client_generators/dependencies/base_client.py
```diff
@@ -70,9 +71,11 @@ def get_data(self, response: httpx.Response) -> dict[str, Any]:
...
-            for key, value in dict_.items()
-        }
+        def maybe_convert_element(path, key, value) -> Union[bool, Tuple[Any, Any]]:
+            if value is None:
+                return False
+            if isinstance(value, BaseModel):
+                return key, value.dict(by_alias=True)
+            else:
+                return key, value
+        return remap(dict_, maybe_convert_element)
```
mat-sop, line 81: I wonder if it is worth introducing a new dependency, isn't it equivalent to something like this?
```py
def _convert_value(self, value):
    if isinstance(value, BaseModel):
        return value.dict(by_alias=True)

    if isinstance(value, list):
        return [self._convert_value(item) for item in value]

    return value


def _convert_dict_to_json_serializable(self, dict_: Dict[str, Any]) -> Dict[str, Any]:
    return {key: self._convert_value(value) for key, value in dict_.items()}
```

mat-sop (commented):
> > Also removes Nones from the dictionary to handle optional inputs.
> 
> This isn't strictly necessary so happy to remove this bit if you would prefer.

If I understand correctly, this change makes it impossible to pass `None`/`null` explicitly?

ariadne_codegen/client_generators/dependencies/base_client.py
strue36: The remap function is a bit of a hangover from another project where there were fewer guarantees on the types that were provided.

If we are guaranteed to only ever have combinations of `list`s and `pydantic.BaseModel`s then yes I think that the behaviour above is equivalent. If we might end up with `dict`s then the extra dependency is adding value.

strue36:
> > > Also removes Nones from the dictionary to handle optional inputs.
> > 
> > 
> > This isn't strictly necessary so happy to remove this bit if you would prefer.
> 
> If I understand correctly, this change makes it impossible to pass `None`/`null` explicitly?

Yes. I had to do a bit of digging today to get my head around how this should work.

The issue I'm having is that I have a lot of optional arguments and optional fields within them.
The current generated code sets every missing value to `None`, and then sends `"field": null`.

Our GraphQL server interprets missing arguments/fields as semantically different to null values and therefore I want to remove them from the json. I spent some time today wondering if this is actually a bug in our server/the framework we are using, but the [GraphQL language spec](http://spec.graphql.org/June2018/#sec-Null-Value) makes it clear that they are semantically different.

I had assumed that `None` representing a missing optional argument would be far more common than explicitly sending `null`, and that this change would result in payloads that are closer to what a user intended (and therefore should be the default behaviour). That might not be the case and if so we can raise an issue for null handling in general and discuss there. I have a similar fix in `base_model.py` for stripping nulls which would raise the same discussion points.

mat-sop:
I'm not a fan of choosing one of these options. I wonder if it is a way to distinguish the default `None` from user-provided `None`. Maybe we should use something different for optional arguments and input fields, (`graphql.Undefined`?) and then remove only this from the payload. @rafalp What do you think?

strue36:
I was going to point you at [this discussion from openapi-python-client](https://github.com/openapi-generators/openapi-python-client/discussions/385#discussioncomment-601623) which explains their "Unset" concept, however that points at [some Pydantic features](https://docs.pydantic.dev/usage/exporting_models/#modeldict) that may save a lot of trouble here.

Should I create a new issue for the null handling as it clearly requires further discussion?

mat-sop:
> Should I create a new issue for the null handling as it clearly requires further discussion?

That would be great, thank you!

tests/client_generators/dependencies/test_async_base_client.py
```diff
@@ -67,6 +68,39 @@ class TestModel2(BaseModel):
...
+
+    assert fake_client.post.called
+    assert len(fake_client.post.mock_calls) == 1
+    call_kwargs = fake_client.post.mock_calls[0].kwargs
+    assert call_kwargs["url"] == "base_url"
+    assert call_kwargs["json"] == {
+        "query": query_str,
+        "variables": {"v1": [[{"a": 1}, {"a": 2}]]},
+    }
+    assert not research(call_kwargs["json"]["variables"], lambda p, k, v: isinstance(v, BaseModel))
```
strue36, line 101: This still has a dependency on boltons. Do you object to that being a dev dependency?

strue36:
I can squash this to remove the commits related to null values before merging.

tests/client_generators/dependencies/test_async_base_client.py
rafalp: Yes, we should have custom utility instead.

rafalp (changes requested):

rafalp:
Deleting nulls from input values is wrong. As mentioned above in Python GraphQL `"key": None` and no key in input have different meaning. This behavior was implemented in Graphene 3 and it caused massive pain to the developers. Adding this behavior to client will make it useless for Saleor GraphQL, which is our primary use case.

Three solutions come to mind:

- Check if there's way to get information from model which fields were initialized from user given values.
- `Undefined` type for no presence of field (there's one in GraphQL utils, but we could re-export it for convenience or define our own).
- Inputs as typed dicts.

strue36:
> Deleting nulls from input values is wrong. As mentioned above in Python GraphQL `"key": None` and no key in input have different meaning. Behavior normalizing those two paths into one was implemented in Graphene 3 and it caused massive pain to the developers. This behavior in client will make it problematic for Saleor GraphQL, which is our primary use case.
> 
> Three solutions come to mind:
> 
> * Check if there's way to get information from model which fields were initialized from user given values.
> * `Undefined` type for no presence of field (there's one in GraphQL utils, but we could re-export it for convenience or define our own).
> * Inputs as typed dicts.

I've raised a new issue for the null discussion (#121), but it looks like Pydantic can handle unset values within the model. We just need a solution for the top level variables which can probably be done with a marker such as `Undefined`

tests/client_generators/dependencies/test_async_base_client.py
strue36: I've replaced this with a check for the specific list structure we have here. It only exists in 2 places. I figured we can write something generic if we end up with more tests that require it.

mat-sop (approved):
Looks good to me 🚀

ariadne_codegen/client_generators/dependencies/async_base_client.py
```diff
@@ -69,12 +69,14 @@ def get_data(self, response: httpx.Response) -> dict[str, Any]:
 
         return cast(dict[str, Any], data)
 
+    def _convert_value(self, value):
```
mat-sop, line 72: I missed it earlier, `_convert_value` misses type annotations
```suggestion
    def _convert_value(self, value: Any) -> Any:
```

mat-sop (approved):

rafalp (approved):

rafalp:
Please rebase on main branch

PR #123, opened 2023-04-05
Change BaseModel to apply parse and serialize methods on every list element

### LINKED ISSUE #104: Custom scalar 'parse' function is applied directly to a List[<scalar>], instead of being mapped to each item in list
hey! another potential bug / opportunity for improvement here - I currently have a custom scalar called `Time` that I would like to deserialise from a string into a `pendulum.time.Time` instance. 

However, the actual return value of the `times` query is a **list** of `Time` instances, not just a single `Time` instance - in this case, I would expect the custom scalar parser function `time_utils.parse_time` to be applied to each element of the `times` array, rather than having the whole array passed into the function once.

**schema.graphql**
```
scalar Time

type Query {
    times: [Time!]!
}
```

**time_utils.py**
```
from pendulum.time import Time
from pendulum.parser import parse

def parse_time(time: str) -> Time:
    return parse(time).time()
```

**pyproject.yaml**
```
[tool.ariadne-codegen]
...
files_to_include = ["time_utils.py"]

[tool.ariadne-codegen.scalars.Time]
type = "Time"
parse = "parse_time"
import = ".time_utils"
```

Happy to submit a PR for this - I'm thinking the `BaseModel.decode_custom_scalars` function could instead be generated as something like:
```
    @validator("*", pre=True)
    def decode_custom_scalars(cls, value: Any, field: ModelField) -> Any:
        decode = SCALARS_PARSE_FUNCTIONS.get(field.type_)
        if decode is None or not callable(decode):
            return value

        if isinstance(value, list):
            return [decode(item) for item in value]

        return decode(value)
```

As a side note, as much as I prefer the `from ... import ...` syntax, for ease of use I would imagine that specifying custom scalar parsing in `pyproject.yaml` in the following way would be nicer,
```
[tool.ariadne-codegen.scalars.Time]
type = "pendulum.time.Time"
parse = ".time_utils.parse_time"
serialize = "str"
```
since the one `time_utils.py` module isn't responsible for exporting all of the types / functions required (including builtins like `str`). Definitely happy to hear why the current way is better though!

mat-sop:
Resolves #104
This pr changes base model to apply parse/serialize method on every list element instead of applying it on the whole list object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Unlocked versions of black, isort, autoflake and dev dependencies
 - Added `remote_schema_verify_ssl` option.
 - Changed how default values for inputs are generated to handle potential cycles.
+- Changed `BaseModel` to apply `parse` and `serialize` methods on every list element.
 
 
 ## 0.4.0 (2023-03-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Unlocked versions of black, isort, autoflake and dev dependencies
 - Added `remote_schema_verify_ssl` option.
 - Changed how default values for inputs are generated to handle potential cycles.
-- Changed `BaseModel` to apply `parse` and `serialize` methods on every list element.
+- Fixed `BaseModel` incorrectly calling `parse` and `serialize` methods on entire list instead of its items for `List[Scalar]`.
 
 
 ## 0.4.0 (2023-03-20)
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Unlocked versions of black, isort, autoflake and dev dependencies
 - Added `remote_schema_verify_ssl` option.
 - Changed how default values for inputs are generated to handle potential cycles.
+- Changed `BaseModel` to apply `parse` and `serialize` methods on every list element.
```
rafalp, line 11:
```suggestion
- Fixed `BaseModel` incorrectly calling `parse` and `serialize` methods on entire list instead of its items for `List[Scalar]`.
```

rafalp (approved):

PR #126, opened 2023-04-06
Change logic how custom scalar imports are generated

### LINKED ISSUE #120: Change custom scalar imports
As discussed in #104

If `parse`/`serialize`/`type`, in scalar section, consists `.` then produced import should be generated by splitting provided string by last `.`, eg. `"ab.cd.CustomScalar"` -> `from ab.cd import CustomScalar`

If scalar section has `import` key then imports should be generated according to the existing strategy with raised deprecation warning

mat-sop:
resolves #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,10 @@ You can provide information about specific scalar by adding section to `pyprojec
 type = "(required) python type name"
 serialize = "function used to serialize scalar"
 parse = "function used to create scalar instance from serialized form"
-import = "module to import from"
 ```
 
-All occurences of `{graphql scalar name}` will be represented as `type`. If provided, `serialize` and `parse` will be used for serialization and deserialization. In all files which use `type`/`serialize`/`parse` there will be added extra import `from {import} import {type}, {serialize}, {parse}`
+All occurences of `{graphql scalar name}` will be represented as `type`. If provided, `serialize` and `parse` will be used for serialization and deserialization.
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -102,10 +102,10 @@ You can provide information about specific scalar by adding section to `pyprojec
 type = "(required) python type name"
 serialize = "function used to serialize scalar"
 parse = "function used to create scalar instance from serialized form"
-import = "module to import from"
 ```
 
-All occurences of `{graphql scalar name}` will be represented as `type`. If provided, `serialize` and `parse` will be used for serialization and deserialization. In all files which use `type`/`serialize`/`parse` there will be added extra import `from {import} import {type}, {serialize}, {parse}`
+All occurences of `{graphql scalar name}` will be represented as `type`. If provided, `serialize` and `parse` will be used for serialization and deserialization.
+If `type`/`serialize`/`parse` contains at least one `.` then string is split by last occurence, first part will be used as module to import from, and second part as type/method name, eg. `type = "custom_scalars.a.ScalarA"` will produce `from custom_scalars.a import ScalarA`.
```
rafalp, line 108:
```suggestion
If `type`/`serialize`/`parse` contains at least one `.` then string will be split by it's last occurrence. First part will be used as module to import from, and second part as type/method name. For example, `type = "custom_scalars.a.ScalarA"` will produce `from custom_scalars.a import ScalarA`.
```

rafalp (approved):

PR #128, opened 2023-04-12
Remove unset arguments and fields from variables payload

### LINKED ISSUE #121: Differentiate between null and unset values.
As discussed in #116

The issue I'm having is that I have a lot of optional arguments and optional fields within them.
The current generated code sets every missing value to `None`, and then sends `"field": null`.

Our GraphQL server interprets missing arguments/fields as semantically different to null values and therefore I want to remove them from the json. I spent some time today wondering if this is actually a bug in our server/the framework we are using, but the [GraphQL language spec](http://spec.graphql.org/June2018/#sec-Null-Value) makes it clear that they are semantically different.

I had assumed that `None` representing a missing optional argument would be far more common than explicitly sending `null`, and that removing any null entry would result in payloads that are closer to what a user intended (and therefore should be the default behaviour). That might not be the case and if so we can raise an issue for null handling in general and discuss there. I have a similar fix in `base_model.py` for stripping nulls which would raise the same discussion points.

> I'm not a fan of choosing one of these options. I wonder if it is a way to distinguish the default `None` from user-provided `None`. Maybe we should use something different for optional arguments and input fields, (`graphql.Undefined`?) and then remove only this from the payload. @rafalp What do you think?

_Originally posted by @mat-sop in https://github.com/mirumee/ariadne-codegen/issues/116#issuecomment-1494145149_

[This discussion from openapi-python-client](https://github.com/openapi-generators/openapi-python-client/discussions/385#discussioncomment-601623) explains their "Unset" concept, however that references [some Pydantic features](https://docs.pydantic.dev/usage/exporting_models/#modeldict) that may save a lot of trouble here.

We need to handle nullability at the GraphQL argument level and within an input type.

mat-sop:
resolves #121 
This pr:
- refactor ClientGenerator to handle the whole client module
- changes the default value of optional arguments from `None` to `UNSET`
- removes unset arguments and input fields from variables payload
- removes duplicated expected schema file from `graphqlschema` test


ariadne_codegen/client_generators/constants.py
```diff
@@ -52,3 +52,6 @@
 
 SCALARS_PARSE_DICT_NAME = "SCALARS_PARSE_FUNCTIONS"
 SCALARS_SERIALIZE_DICT_NAME = "SCALARS_SERIALIZE_FUNCTIONS"
+
+UNSET = "UNSET"
```
rafalp, line 56: This should be `class Unset:` because current implementation prevents use of  `"UNSET"` value in inputs.

rafalp (changes requested):

ariadne_codegen/client_generators/constants.py
mat-sop: This constant is used by generator to keep track of variable's name with constant, I changed it to `UNSET_NAME`. `UNSET` used by the generated client is declared [here](https://github.com/mirumee/ariadne-codegen/blob/unset_values/ariadne_codegen/client_generators/dependencies/base_model.py#L15)
```py
class UnsetType:
    def __bool__(self) -> bool:
        return False


UNSET = UnsetType()
```

rafalp: Oh okay, that confused me a bit here.

Issue #100, opened 2023-03-18
Multiple failures with a Hasura-generated graphql schema


ataibarkai:
This project has a tone of promise. Architecturally it looks far, far better than any other python graphql client out there.

However I'm running into multiple issues when using a [Hasura](https://hasura.io/)-generated graphql schema (attached).
[hasura_schema.garphql.txt](https://github.com/mirumee/ariadne-codegen/files/11007098/hasura_schema.garphql.txt)
[the attachment postfix is `.graphql.txt` - so that GitHub would allow me to upload it. Just remove the `.txt` postfix].

A few problems I've run into:

1. Hasura's root Query type is called `query_root`, not `Query` - but `ariadne-codegen` appears to assume it's always called `Query` and fail otherwise (`no property called "..." on type "Query"`...). The GraphQL spec doesn't seem to require a specific name, so would be great to be able to customize this in ariadne-codegen. **I manually renamed `query_root` to `Query`  (in 2 places in the schema) to proceed.**

2. You get into an infinite recursion loop in `_get_dependant_names`. I didn't look at the codebase in detail to see if this fix fully makes sense, but here's the change I made to the relevant function. **I added `pre_callers` argument to avoid recursive cycles:**

```
def _get_dependant_names(self, name: str, pre_callers: List[str] = []) -> List[str]:
    pre_callers = pre_callers.copy()
    
    if name in pre_callers:
        return []
    
    pre_callers.append(name)
    result = []
    
    for dependency_name in self._dependencies[name]:
        if dependency_name != name:
            result.extend(self._get_dependant_names(dependency_name, pre_callers))
            result.append(dependency_name)
            
    return result
```

3. After fixing those errors I got a new one:
`black.parsing.InvalidInput: Cannot parse: 38:4:     in: Optional[List[bool]] = Field(alias='_in')`. **Gave up here**.

mat-sop:
Thank you for reporting all of the issues, I looked into them and have some thoughts:

1. We should handle other names for query/mutation types without specifying them in the config. I created a separate issue for this bug #101 .


2. It Looks like the issue is caused by mutually dependent inputs, eg. 
```gql
input person_entities_bool_exp {
  ...
  person_entity_clips: person_entity_clips_bool_exp
  ...
}
input person_entity_clips_bool_exp {
  ...
  person_entity: person_entities_bool_exp
  ...
}
```
so your fix will work (at least for your schema) but I think it needs further investigation regarding other cases and fields' default values.


[...]

PR #129, opened 2023-04-12
Add process_name hook

mat-sop:
This pr adds `process_name` hook to allow further customization of generated fields/arguments/method names. Requested [here](https://github.com/mirumee/ariadne-codegen/issues/100#issuecomment-1499754435)

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -287,7 +287,7 @@ Hook executed on generation of init code. Result is used as content of `__init__
 ### process_name
 
 ```py
-def process_name(self, name: str) -> str:
+def process_name(self, name: str, node: Optional[Node] = None) -> str:
 ```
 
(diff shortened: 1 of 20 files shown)


PLUGINS.md
```diff
@@ -284,6 +284,14 @@ def generate_init_code(self, generated_code: str) -> str:
 
 Hook executed on generation of init code. Result is used as content of `__init__.py`.
 
+### process_name
+
+```py
+def process_name(self, name: str) -> str:
```
rafalp, line 290: Can we provide more context about the item which name is being processed by this hook, eg. an AST node? This would let plugin authors use different logic for different types.

mat-sop: Done

rafalp (approved):

PR #133, opened 2023-04-14
Draft: Generate subscriptions as async generators

### LINKED ISSUE #65: Support GraphQL subscriptions
Subscriptions could be async generators consuming the GraphQL websockets.

mat-sop:
resolves #65 
Example of generated client:
```py
class Client(AsyncBaseClient):
    async def get_counter(self) -> AsyncIterator[GetCounter]:
        query = gql(
            """
            subscription getCounter {
              counter
            }
            """
        )
        variables: dict[str, object] = {}
        async for data in self.execute_ws(query=query, variables=variables):
            yield GetCounter.parse_obj(data)
```


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## UNRELEASED
+
+- Added generation of subscriptions as async generators.
```
rafalp, line 5:
```suggestion
- Added support for subscriptions as async generators.
```

ariadne_codegen/client_generators/client.py
```diff
@@ -115,28 +129,39 @@ def add_method(
         arguments, arguments_dict = self.arguments_generator.generate(
             definition.variable_definitions
         )
-        method_def = (
-            self._generate_async_method(
+        if definition.operation == OperationType.SUBSCRIPTION:
+            if not async_:
+                raise NotSupported("Only async client supports subscriptions.")
```
rafalp, line 134:
```suggestion
                raise NotSupported("Subscriptions are only available when using async client.")
```

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
   "toml~=0.10",
   "httpx~=0.23",
   "pydantic~=1.10",
+  "websockets~=11.0",
```
rafalp, line 28: Make this dependency opt-in, eg. `pip install ariadne-codegen[subscriptions]`

rafalp (approved):

pyproject.toml
mat-sop: Hm I'm not sure if it's the right way. `AsyncBaseClient` still depends on `websockets` even when no subscription was generated.

rafalp: It should handle `websockets` not being installed. Usual approach is handling the import error and importing placeholder object that implements the subset of API but raises not implemented error from its methods.

EXAMPLE.md
```diff
@@ -325,9 +347,9 @@ class Color(str, Enum):
     YELLOW = "YELLOW"
 ```
 
-### Query/mutation types
+### Query/mutation/subscription types
 
-For every provided query/mutation there is generated file, that contains models which correspond to return type of operation. File name is generated by converting query/mutation name to snake case. Root models has the same name as query/mutation, and depend classes use it as prefix in their names.
+For every provided query/mutation/subscription there is generated file, that contains models which correspond to return type of operation. File name is generated by converting operation name to snake case. Root models has the same name as query/mutation/subscription, and depend classes use it as prefix in their names.
```
rafalp, line 352:
```suggestion
For every provided query/mutation/subscription there is a generated file that contains models which correspond to return type of operation. File name is generated by converting operation name to snake case. Root models has the same name as query/mutation/subscription, and depend classes use it as prefix in their names.
```

tests/main/clients/remote_schema/expected_client/async_base_client.py
```diff
@@ -7,25 +9,61 @@
...
+    WebSocketClientProtocol = Any  # type: ignore
+    Data = Any  # type: ignore
+    Origin = Any  # type: ignore
+    Subprotocol = Any  # type: ignore
+
+
 Self = TypeVar("Self", bound="AsyncBaseClient")
 
+GQL_SUBPROTOCOL = "graphql-transport-ws"
+GQL_CONNECTION_INIT = "connection_init"
```
rafalp, line 37: We could group message types under an Enum

rafalp (commented):
We should mention in readme that websockets are optional dependency required for subscriptions.

rafalp (approved):

PR #137, opened 2023-04-24
Add __main__ to allow executing via python -m ariadne_codegen

huonw:
This adds a `__main__.py` file, which acts as an entry point when the `ariadne_codegen` module is executed directly, such as via `python -m ariadne_codegen`. This is in addition to the existing console script (i.e. `ariadne-codegen` directly still works).

Docs: https://docs.python.org/3/library/__main__.html#main-py-in-python-packages

Executing a module directly like this is widely supported by other python tools, and is often even recommended, because, AIUI, `python -m ...` will execute within `python`'s venv, while a console script wrapper may not and thus pick up an unexpected version. For instance, other tools like `pytest` and `mypy` support these parallel models of execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ It's available as `ariadne-codegen` command and reads configuration from the `py
 $ ariadne-codegen
 ```
 
+It can also be run as `python -m ariadne_codegen`.
+
 
 ## Features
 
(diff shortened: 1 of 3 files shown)


rafalp:
LGTM. Could you please also add small note to the readme that `python -m ariadne_codegen` is also supported?

huonw:
Done!

mat-sop (approved):

huonw:
Thank you!

Issue #66, opened 2023-02-01
Shorthand queries and mutations where result returns single field.


rafalp:
Taking a query like this:

```graphql
query GetAuthenticatedUser {
  me {
    id
    username
  }
}
```

Codegen would generate a client method that would return `GetAuthenticatedUserUser` directly from the `me` field, so there's no need for:

```python
result = client.get_authenticated_user()
auth_user = result.me
```

Instead we would get:

```python
auth_user = client.get_authenticated_user()
```

We could have an setting like `unwrap_single_fields=false` for enabling this behavior.

EDjur:
Hi! Thanks for taking time to think about this in such a short notice.

When exploring this initially I also gleaned over `generate_result_class` if I could somehow write a plugin that could help me to do what I want. I think your second approach makes a lot of sense and is non-invasive.

Do you see that being an "official" plugin that would sit in this repo or up to users to write themselves? It seems a relatively common use case so I guess a more official plugin makes sense?

mat-sop:
In my opinion we can treat this plugin as "official" and include it in repo. Maybe using `unwrap_single_fiels` we can "hide" it (but still document how this flag works), in [settings](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/settings.py#L59) for example:
```py
if self.unwrap_single_fiels:
    self.plugins.insert(0, "ariadne_codegen.ShorterResultsPlugin")
```

@rafalp What do you think about including this plugin in repo and still using `unwrap_single_fiels` flag?

rafalp:
There are features we want to do out of the box. But if those features are contributed to us as a plugin, I feel that we should treat them as a plugins and don't shortcut them under dedicated settings. Like how Ariadne and Django does some extra features as extensions and keeps them under `.contrib` package.

Doing it like this we get extra bonus of having real plugins in package for plugin devs to look  to for example and inspirations.

mat-sop:
Resolved by #141 @bombsimon Thank you for contributing :heart:

PR #141, opened 2023-05-10
feat: Add plugin to unwrap single field classes

bombsimon:
Add a plugin that can be enabled by setting `unwrap_single_fields` to true in the settings to automatically return the inner value if a return type from a generated function only has a single method.

Ref: #66

---

This PR isn't in draft because the feature is complete but I would love some feedback on the code and approach. Even by following the suggestions in #66 there are probably things that could be solved different. I added a few TODOs/questions inline.

Also this PR doesn't yet include any tests but I saw you had a nice setup with `expected_client` for general client generated code so if this approach is resonable I'd be happy to ad some tests with the plugin enabled to ensure it's tested.


ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+        change the return type.
+        """
+        if len(method_def.body) < 1:
+            return super().generate_client_method(method_def)
+
+        # TODO: Can we assume that the return statement is always the last
+        # statement?
+        return_stmt = method_def.body[-1]
+        if not isinstance(return_stmt, ast.Return):
+            return super().generate_client_method(method_def)
```
bombsimon, line 110: This will obviously break if the codegen introduces a new path for returns but right now it seems a bit unlikely. I guess the alternative would be to store all return statements in a list, filter out those that's not for the generated class (f.ex. if there are `None` for optionals) and then update all.

I didn't do that because such refactor could change this as well instead of doing pre-mature optimization but maybe there's a need for it if someone else chains plugins and change the ast with a custom plugin introducing this?

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+        # other things not generated this iteration so ignore those.
+        # TODO: Can we ensure this in a better way?
+        try:
+            single_field_class = ast.literal_eval(single_field_class)
+        except ValueError:
+            # Not a literal
+            pass
+
+        if single_field_class not in self.class_dict:
+            return super().generate_client_method(method_def)
```
bombsimon, line 156: I'm not sure about this but I need to trim the class I get from the ast since it's quoted. However, we can also get types like `typing.Any` which will make `literal_eval` fail.

Further this branch will not enable the plugin for non generated types (as per the docs) but maybe there will be use cases where we want to do this even for `typing.Any`? I think custom scalars work like this? If so would the client method just return `typing.Any`?

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+
+        method_def.body[-1] = ast.Attribute(
+            value=return_stmt,
+            attr=single_field_name,
+        )
+
+        # After we change the type we also need to import it from the client so
+        # add the extra import as needed.
+        if method_def.name not in self.extended_imports:
+            self.extended_imports[method_def.name] = set()
```
bombsimon, line 175: This will break if the method name isn't the same as the filename but I didn't find any other way to figure out which file would need to be updated. Are there better alternatives to use as a key to ensure we can pick this up in `generate_client_module`?

bombsimon:
Thank you for the very quick and through review! I'll try to get back to addressing them as soon as possible.

ariadne_codegen/plugins/base.py
```diff
@@ -19,6 +19,8 @@ class Plugin:
     def __init__(self, schema: GraphQLSchema, config_dict: Dict) -> None:
         self.schema = schema
         self.config_dict = config_dict
+        self.class_dict = {}
+        self.extended_imports = {}
```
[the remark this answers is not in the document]
bombsimon: Oops, of course!

ariadne_codegen/plugins/shorter_results.py
bombsimon: Cool! Kept as is because I want to check lengt of generator and see if we're a return or async for statement!

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+        # represented as an `ast.Subscript`.
+        # We need to update the proper return type to be optional as well so we
+        # store that. The ast in the annotation isn't an evaluated node so we
+        # can't return that as is.
+        if isinstance(single_field.annotation, ast.Name):
+            is_optional = False
+            single_field_class = single_field.annotation.id
+        elif isinstance(single_field.annotation, ast.Subscript):
+            is_optional = True
+            single_field_class = single_field.annotation.slice.id
```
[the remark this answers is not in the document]
bombsimon: Thanks for the info! This is my first time working with python ast so I'll miss more like this.

I removed this and kept the annotation as is which works on all types, just traversing to eval.

ariadne_codegen/plugins/shorter_results.py
bombsimon: Makes sense, removed this and added `str` in test!

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+            method_def.returns = ast.Subscript(
+                value=ast.Name(id="Optional"), slice=ast.Name(id=single_field_class)
+            )
+        else:
+            method_def.returns = ast.Name(id=single_field_class)
+
+        method_def.body[-1] = ast.Attribute(
+            value=return_stmt,
+            attr=single_field_name,
+        )
```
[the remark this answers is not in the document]
bombsimon: Changed to return and async for respectively, created with existing codegen

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,179 @@
...
+
+        return_class = method_def.returns.id
+        if return_class not in self.class_dict:
+            return super().generate_client_method(method_def)
+
+        return_class_ast = self.class_dict[return_class]
+
+        # If we're not a single method data class move on!
+        if len(return_class_ast.body) != 1:
+            return super().generate_client_method(method_def)
```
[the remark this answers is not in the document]
bombsimon: Right... I didn't address this at all yet, need to dig some more. Do you have any suggestions or ideas on how to tackle this?

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,283 @@
...
+
+        if not isinstance(method_def.body[-1].body[0], ast.Expr):
+            return super().generate_client_method(method_def)
+
+        if not isinstance(method_def.body[-1].body[0].value, ast.Yield):
+            return super().generate_client_method(method_def)
+
+        previous_yield_value = method_def.body[-1].body[0].value.value
+        if previous_yield_value is None:
+            return super().generate_client_method(method_def)
```
bombsimon, line 171: This is to make the type checker happy but it feels s bit verbose. I used similar patterns in some other places. 

Is this idiomatic enough? Can I simplify this when working with multiple `ast.stmt`?

ariadne_codegen/plugins/shorter_results.py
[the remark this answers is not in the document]
bombsimon: Ok I did a quick pass just to get an idea of this. The first case is straight forward, if we only have a single field, check that we inherit from `BaseModel` (I guess there are no other classes to consider?). If we inherit from anything else we don't unwrap the field. We could traverse the inherited class (if we somehow get the `ClassDef`) but I can't think of a scenario where we would inherit from a class _without_ fields.

For the second example that's a bit trickier because we never see `TestQuery` in `generate_result_class` so it won't get added to our dict and thus I won't be able to get what class it inherits from. I see we get the code for the class in `generate_result_types_code` but that one is called _after_ `generate_client_method` so even if I parse the code to get all the generated classes it will be too late. Also we don't have the ast for the fragment either but that's exposed in `generate_fragments_module` which is also called _after_ `generate_client_method` so not sure how much that will help.

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+
+    if not isinstance(return_class_ast.bases[0], ast.Name):
+        return None
+
+    if not return_class_ast.bases[0].id == "BaseModel":
+        return None
+
+    single_field = return_class_ast.body[0]
+    if single_field is None:
+        return None
```
mat-sop, line 259: I think we should "unpack" fragments, for example:
```py
def _get_all_fields(
    class_def: ast.ClassDef, class_dict: Dict[str, ast.ClassDef]
) -> List[ast.AnnAssign]:
    fields = []
    for base in class_def.bases:
        if not isinstance(base, ast.Name):
            continue
        fields.extend(_get_all_fields(class_dict[base.id], class_dict=class_dict))

    for field in class_def.body:
        if isinstance(field, ast.AnnAssign):
            fields.append(field)

    return fields
```
And then:
```suggestion
    fields = _get_all_fields(class_dict[current_return_class], class_dict)
    if not fields or len(fields)!= 0:
        return None
    single_field = fields[0]
```

I see that for now this might be difficult, as `generate_result_class` for fragments is called after `generate_client_method` but I just created pr to fix this.

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def _update_node(node: ast.expr) -> tuple[ast.expr, str]:
+    """
+    Recurse down a node to finner the inner ast.Name. Once found, evaluate the
+    potential literal so it gets unquoted and return the inner name.
+    """
+    if isinstance(node, ast.Name):
+        try:
+            node.id = ast.literal_eval(node.id)
```
mat-sop, line 282: This will also update fields annotation, either we should "rewrite" whole annotation with modified values or use original one without changing it. For example (from test you added) `ListTypeAListOptionalTypeA` is no longer quoted:
```py
list_optional_type_a: List[Optional[ListTypeAListOptionalTypeA]] = Field(
        alias="listOptionalTypeA"
    )
```

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+    generate_name,
+    generate_return,
+    generate_subscript,
+    generate_yield,
+)
+
+from ariadne_codegen.plugins.base import (
+    Plugin,
+    SelectionSetNode,
+)
```
mat-sop, line 58:
```suggestion
from graphql import GraphQLSchema, ExecutableDefinitionNode, SelectionSetNode
from ariadne_codegen.codegen import (
    generate_async_for,
    generate_attribute,
    generate_expr,
    generate_name,
    generate_return,
    generate_subscript,
    generate_yield,
)

from ariadne_codegen.plugins.base import Plugin
```

ariadne_codegen/settings.py
```diff
@@ -53,6 +53,7 @@ class ClientSettings(BaseSettings):
     include_comments: bool = True
     convert_to_snake_case: bool = True
     async_client: bool = True
+    unwrap_single_fields: bool = False
```
mat-sop, line 56: As said [here](https://github.com/mirumee/ariadne-codegen/issues/66#issuecomment-1545578003) let's remove this setting and add information about plugin to `README.md`

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+    if isinstance(node, ast.Name):
+        try:
+            node.id = ast.literal_eval(node.id)
+        except ValueError:
+            # Not a literal
+            pass
+
+        return node, node.id
+
+    if isinstance(node, ast.Subscript):
```
mat-sop, line 289: We also need to handle `ast.Tuple` in case of Union or Interface, eg.
```gql
type Query {
  queryI: Interface!
}

interface Interface {
  id: String!
}

type TypeA implements Interface {
  id: String!
  fieldA: String!
}

type TypeB implements Interface {
  id: String!
  fieldB: String!
}
```
```gql
query TestQuery {
  queryI {
    id
    ... on TypeA {
      fieldA
    }
    ... on TypeB {
      fieldB
    }
  }
}
```
This query will generate follwing type with 1 field with Union as annotation
```py
class TestQuery(BaseModel):
    query_i: Union[
        "TestQueryQueryIInterface", "TestQueryQueryITypeA", "TestQueryQueryITypeB"
    ] = Field(alias="queryI")
``` 

[Annotation](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/client_generators/types.py#L15) can be represented as `ast.Name` or `ast.Subscript`, but subscript's [slice](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/client_generators/types.py#L16) can be represented as `ast.Name`, `ast.Subscript` or `ast.Tuple`

ariadne_codegen/plugins/shorter_results.py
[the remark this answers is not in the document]
mat-sop: Sorry, I missed this comment, I addressed this issue [here](https://github.com/mirumee/ariadne-codegen/pull/141#discussion_r1193909475)

ariadne_codegen/plugins/shorter_results.py
bombsimon: Split this to a separate method

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        if len(method_def.body) < 1:
+            return super().generate_client_method(method_def)
+
+        last_stmt = method_def.body[-1]
+
+        if isinstance(last_stmt, ast.Return):
+            return self._generate_sync_client_method(method_def, last_stmt)
+        elif isinstance(last_stmt, ast.AsyncFor):
+            return self._generate_async_client_method(method_def, last_stmt)
```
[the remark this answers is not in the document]
bombsimon: Renamed! `_generate_subscription_client_method` and `_generate_query_and_mutation_client_method`.

ariadne_codegen/plugins/shorter_results.py
```diff
@@ -0,0 +1,294 @@
...
+            method_def.returns.slice.id, self.class_dict
+        )
+        if node_and_class is None:
+            return super().generate_client_method(method_def)
+
+        yield_node, single_field_class, single_field_return_class = node_and_class
+
+        # Update the return type to be the inner field and ensure we reference
+        # the single field before returning.
+        method_def.returns = generate_subscript(
```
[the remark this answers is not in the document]
bombsimon: Oof, my bad! Nice catch, thanks!

ariadne_codegen/plugins/shorter_results.py
bombsimon: Awesome suggestion, makes sense! Added this but there are still some issues.

1. We return if we don't have the class in our class dict. This is what we want for `BaseClass` f.ex. but is there any risk that this could be a class with actual fields?
2. Even with this I still don't generate fragments first. It's _called_ by the fragments, but that's after I generated the method. I tried adding this piece:

```python
    def generate_fragments_module(
        self,
        module: ast.Module,
        fragments_definitions: Dict[str, FragmentDefinitionNode],
    ) -> ast.Module:
        class_defs = list(filter(lambda x: isinstance(x, ast.ClassDef), module.body))
        for class_def in class_defs:
            if not isinstance(class_def, ast.ClassDef):
                continue

            self.class_dict[class_def.name] = class_def

        return module
```

But that gets called _after_ `generate_client_method` still. 🤔

ariadne_codegen/settings.py
bombsimon: Removed settings and added section to readme. Not sure how you want to structure this, I just added it in the root with a short summary of what it does, linking to it for further docs.

ariadne_codegen/plugins/shorter_results.py
bombsimon: Now making a copy of the annotation and updating the one we're using in the client but keeping it quoted for the original annotation.

ariadne_codegen/plugins/shorter_results.py
mat-sop: 1. Generated model can only inherit from `BaseModel` (which we can safely assume doesn't have any fields we should care about) or fragments mixins. Each fragment mixin can only inherit from other fragments or `BaseModel`. I see that in my  `_get_all_fields` I didn't handle the fact that `BaseModel` is not in `class_dict`.
```py
def _get_all_fields(
    class_def: ast.ClassDef, class_dict: Dict[str, ast.ClassDef]
) -> List[ast.AnnAssign]:
    fields = []
    for base in class_def.bases:
        if not isinstance(base, ast.Name) or base.id not in class_dict:
            continue
        fields.extend(_get_all_fields(class_dict[base.id], class_dict=class_dict))

    for field in class_def.body:
        if isinstance(field, ast.AnnAssign):
            fields.append(field)

    return fields
```
2. Oh, you are right. But now we can modify methods in `generate_client_module` which is called after all `generate_result_class`, including models for fragments. Then we don't need `generate_fragments_module`, and it could look like this:
```py
    def generate_client_module(self, module: ast.Module) -> ast.Module:
        client_def = next(
            filter(lambda o: isinstance(o, ast.ClassDef), module.body), None
        )
        if not client_def:
            return super().generate_client_module(module)

        for method_def in [
            m
            for m in client_def.body
            if isinstance(m, (ast.FunctionDef, ast.AsyncFunctionDef))
        ]:
            self._modify_method_def(method_def)

        # imports stuff which is already there
        return super().generate_client_module(module)
    
    def _modift_method_def(self, method_def: Union[ast.FunctionDef, ast.AsyncFunctionDef]):
        # stuff which currently is in generate_client_method
```
This would also require changing how current validation works, so 1 "invalid" method wouldn't stop us from updating the rest of them. What do you think about this approach?

mat-sop: EDIT: I think we wrote a response at almost the same time 😅  Plugin's instance is created once at the beginning of  script execution

bombsimon: Figured out why. If we don't specify any fields in `TestQuery` that class will be empty like you wrote before:

```python
class TestQuery(TestFragment):
    pass
```

When a class is empty it will never be passed to `generate_result_class` so we won't add it to our dict and thus we can't see what class it extends. I'll see if I can figure out how we can still pass the class just for plugin knowledge.

mat-sop: Oh, that shouldn't be the case. I'll look into this, maybe I broke this when I pushed changes related to fragments.

bombsimon: Looks to be related to `ariadne_codegen/client_generators/result_types.py` and `_parse_type_definition`. We only call `generate_result_class` when iterating over `resolved_selection_set`. One solution to fix this would be to call the plugin class if the selection set is empty:

```diff
diff --git a/ariadne_codegen/client_generators/result_types.py b/ariadne_codegen/client_generators/result_types.py
index a3f895d..9bca05e 100644
--- a/ariadne_codegen/client_generators/result_types.py
+++ b/ariadne_codegen/client_generators/result_types.py
@@ -256,6 +256,13 @@ class ResultTypesGenerator:
                     selection_set=selection_set,
                 )
 
+        if len(resolved_selection_set) == 0 and self.plugin_manager:
+            class_def = self.plugin_manager.generate_result_class(
+                class_def,
+                operation_definition=self.operation_definition,
+                selection_set=selection_set,
+            )
+
         if not class_def.body:
             class_def.body.append(generate_pass())
 ```

But maybe that's a bad approach.

Comment (mat-sop) [reply ariadne_codegen/plugins/shorter_results.py]: Yeah, I found the same issue, turns out it was broken even before fragments change 😓 I created [pr](https://github.com/mirumee/ariadne-codegen/pull/147) to fix this

**Review by rafalp — CHANGES_REQUESTED** [review]

File: README.md [hunk 16]
```diff
@@ -84,6 +84,12 @@ Optional settings:
 
 Ariadne Codegen implements a plugin system that enables further customization and fine-tuning of generated Python code. It’s documentation is available separately in the [PLUGINS.md](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md) file.
 
+### Built-in plugins
+
+There are some built-in plugins that ships with `ariadne_codegen`:
+
+* [`ariadne_codegen.ShorterResultsPlugin`](plugins/shorter_results.py) - Unwrap
```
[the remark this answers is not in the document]
bombsimon: Added your suggested quote for now!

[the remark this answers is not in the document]
rafalp: > Sorry, I saw your comment about that but didn't know if you meant in this project or a separate one. I'll move this inside the current project!

Same project. I just want to keep official plugins separate from rest of codebase under specific namespace, like how Ariadne or Django does it.

> I also wanted to avoid over documenting it in the main README,

Don't worry about that. We can always change it later when we become unhappy with it :D

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,369 @@
...
+        self, method_def: Union[ast.FunctionDef, ast.AsyncFunctionDef]
+    ) -> Union[ast.FunctionDef, ast.AsyncFunctionDef]:
+        """
+        Change the method generated in the client to call the inner field and
+        change the return type. We do this here instead of
+        `generate_client_method` to ensure we have a map between all our classes
+        when updating the method def.
+        """
+        if len(method_def.body) < 1:
+            return super().generate_client_method(method_def)
```
mat-sop, line 142: With current approach `generate_client_method` is no longer used, so we should switch to something else as fallback. We don't use output of this method so imo we could just use `return None` instead.

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        client_def = next(
+            filter(lambda o: isinstance(o, ast.ClassDef), module.body), None
+        )
+        if not client_def:
+            return super().generate_client_module(module)
+
+        for method_def in [
+            m
+            for m in client_def.body
```
mat-sop, line 106: Mypy fails on this line:
`ariadne_codegen/contrib/shorter_results.py:106: error: "stmt" has no attribute "body"  [attr-defined]`

ariadne_codegen/contrib/shorter_results.py
bombsimon: Ah, returning `None` in all the paths from non inherited classes now instead!

ariadne_codegen/contrib/shorter_results.py
bombsimon: Fixed by asserting that `client_def` is `ast.ClassDef`

bombsimon:
Sorry, saw that the inports are faulty, will fix!

README.md
```diff
@@ -84,6 +84,17 @@ Optional settings:
 
 Ariadne Codegen implements a plugin system that enables further customization and fine-tuning of generated Python code. It’s documentation is available separately in the [PLUGINS.md](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md) file.
 
+### Standard plugins
```
rafalp, line 87: We use two blank lines before new section headers:

```suggestion

### Standard plugins
```

README.md
```diff
@@ -84,6 +84,17 @@ Optional settings:
...
+
+Ariadne Codegen ships with optional plugins importable from the
+`ariadne_codegen.contrib` package:
+
+* [`ariadne_codegen.contrib.shorter_results.ShorterResultsPlugin`](ariadne_codegen/contrib/shorter_results.py)
+  \- This plugin processes generated client methods for operations where only
+  single top field is requested, so they return this field's value directly
+  instead of operation's result type. For example get_user method generated for
+  query `GetUser() { user(...) { ... }}` will return value of user field
+  directly instead of `GetUserResult`.
```
rafalp, line 97:
```suggestion
  directly instead of `GetUserResult`.

```

README.md
```diff
@@ -84,6 +84,17 @@ Optional settings:
 
 Ariadne Codegen implements a plugin system that enables further customization and fine-tuning of generated Python code. It’s documentation is available separately in the [PLUGINS.md](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md) file.
 
+### Standard plugins
+
+Ariadne Codegen ships with optional plugins importable from the
+`ariadne_codegen.contrib` package:
+
+* [`ariadne_codegen.contrib.shorter_results.ShorterResultsPlugin`](ariadne_codegen/contrib/shorter_results.py)
+  \- This plugin processes generated client methods for operations where only
```
rafalp, line 93:
```suggestion
  \- This plugin processes generated client methods for operations where only
```

Please don't do machinations like that in markdown. ;)

README.md
```diff
@@ -84,6 +84,17 @@ Optional settings:
 
 Ariadne Codegen implements a plugin system that enables further customization and fine-tuning of generated Python code. It’s documentation is available separately in the [PLUGINS.md](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md) file.
 
+### Standard plugins
+
+Ariadne Codegen ships with optional plugins importable from the
```
rafalp, line 89: Please leave text wrapping to user preference. We don't do hard linebreaks in  our markdown.

rafalp (changes requested):
Left some notes about README formatting.

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+        me: Optional["GetAuthenticatedUserMe"]
+
+When later using the client to call `get_authenticated_user` the return type
+would most of the time immediately be expanded to get the inner
+`GetAuthenticatedUserMe`:
+
+    me = (await client.get_authenticated_user).me
+
+By enabling this plugin the return type of the generated
+`get_authenticated_user` would instead directly return a
```
rafalp, line 27:
```suggestion
`get_authenticated_user` will instead directly return a
```

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+`GetAuthenticatedUserMe` so the caller could use:
+
+    me = await client.get_authenticated_user
+
+This plugin can be enabled by either adding the plugin in the settings:
```
rafalp, line 32:
```suggestion
This plugin can be enabled in the settings:
```

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+
+    return ast.expr(), []
+
+
+def _get_all_fields(
+    class_def: ast.ClassDef, class_dict: Dict[str, ast.ClassDef]
+) -> List[ast.AnnAssign]:
+    """
+    Recursively get all fields from all inherited classes to figure out the
+    total amount of fields.
```
rafalp, line 346:
```suggestion
    total number of fields.
```

rafalp (changes requested):
More comments

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+            self._generate_subscription_client_method(method_def, last_stmt)
+
+        return None
+
+    def _generate_subscription_client_method(
+        self,
+        method_def: Union[ast.FunctionDef, ast.AsyncFunctionDef],
+        async_for_stmt: ast.AsyncFor,
+    ):
+        """
```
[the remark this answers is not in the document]
bombsimon: Fixed!

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+    # method where we import the actual types.
+    annotations = deepcopy(single_field.annotation)
+    return_node, return_classes = _update_node(annotations)
+
+    return return_node, return_classes, single_field.target.id
+
+
+def _update_node(node: ast.expr) -> tuple[ast.expr, List[str]]:
+    """
+    Recurse down a node to finner the inner ast.Name. Once found, evaluate the
```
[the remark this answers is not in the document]
bombsimon: Uhh, i don't remember, I guess "find" and "inner" in the same word. Rephrased to "Walk down a node to find the inner ast.Name".

ariadne_codegen/contrib/shorter_results.py
```diff
@@ -0,0 +1,360 @@
...
+    plugins = ["ariadne_codegen.contrib.shorter_results.ShorterResultsPlugin"]
+"""
+
+import ast
+from copy import deepcopy
+from typing import Dict, List, Optional, Union
+
+from graphql import ExecutableDefinitionNode, GraphQLSchema, SelectionSetNode
+
+from ariadne_codegen.codegen import (
```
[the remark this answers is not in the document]
bombsimon: Fixed!

README.md
bombsimon: Do you mean escaping the `-`? I removed the wrapping so no longer escaped

README.md
bombsimon: Sorry, my bad. I use `markdownlint` and my editor does it for me, didn't think about the rest of the file. I changed this to be a single super long line now

bombsimon:
Thanks for the very thorough review and for the help with suggestions! 🙇 I applied your suggestions and addressed the rest

README.md
rafalp: This line still has hard linebreak in it.

bombsimon: 🤦 My bad, sorry for being sloppy. Fixed now!

rafalp (approved):

PR #142, opened 2023-05-10
Demonstrate bug with nested inline fragments

EDjur:
Hi 👋 

First of all, thanks again for this project. Secondly, don't merge as this is just a demonstration of the issue.

I ran into what looks like a bug when expanding inline fragments. I added a test case to demonstrate that if you have a query that directly expands a fragment, the generated client will look different than the `inline_fragments` example.

Given the following schema
```
type Query {
  queryI: Interface!
}

interface Interface {
  id: String!
}

type TypeA implements Interface {
  id: String!
  fieldA: String!
}

type TypeB implements Interface {
  id: String!
  fieldB: String!
}
```

And the following queries:
```
fragment typeAandB on Interface {
  id
  ... on TypeA {
    fieldA
  }
  ... on TypeB {
    fieldB
  }
}

query InterfaceA {
  queryI {
    ...typeAandB
  }
}
```

Then the generated classes will be different than if you expand it like so:
```
query InterfaceA {
  queryI {
    id
    ... on TypeA {
      fieldA
    }
    ... on TypeB {
      fieldB
    }
  }
}
```

This is the result of generating the client when having a query that directly expands a fragment like in the example:
```
from pydantic import Field

from .base_model import BaseModel


class InterfaceA(BaseModel):
    query_i: "InterfaceAQueryI" = Field(alias="queryI")


class InterfaceAQueryI(BaseModel):
    id: str


InterfaceA.update_forward_refs()
InterfaceAQueryI.update_forward_refs()
```

Whereas I would expect it to look the same as in the `inline_fragments` example


EDjur:
cc @mat-sop @rafalp

rafalp:
Please don't @at at us - we have enabled notifications for new PR's and issues for Ariadne repos ;)

mat-sop:
@EDjur Looks like inline fragments inside of fragments are ignored and "skipped", I will investigate this.

EDjur:
Thank you - and sorry again for the ping. Really appreciate the work you've done with this project 🙏

bombsimon:
This seems to be fixed after landing #143. If I rebase on latest main, this branch now generates this for `interface_a.py`:

```python
from typing import Union

from pydantic import Field

from .base_model import BaseModel


class InterfaceA(BaseModel):
    query_i: Union[
        "InterfaceAQueryIInterface", "InterfaceAQueryITypeA", "InterfaceAQueryITypeB"
    ] = Field(alias="queryI")


class InterfaceAQueryIInterface(BaseModel):
    __typename__: str = Field(alias="__typename")
    id: str


class InterfaceAQueryITypeA(BaseModel):
    __typename__: str = Field(alias="__typename")
    id: str
    field_a: str = Field(alias="fieldA")


class InterfaceAQueryITypeB(BaseModel):
    __typename__: str = Field(alias="__typename")
    id: str
    field_b: str = Field(alias="fieldB")


InterfaceA.update_forward_refs()
InterfaceAQueryIInterface.update_forward_refs()
InterfaceAQueryITypeA.update_forward_refs()
InterfaceAQueryITypeB.update_forward_refs()
```

Can you confirm @EDjur?

mat-sop:
@bombsimon From the result it looks like you used query without fragment, which works as expected. When you try query with `...typeAandB` it's still broken.

bombsimon:
> @bombsimon From the result it looks like you used query without fragment, which works as expected. When you try query with `...typeAandB` it's still broken.

Oh, sorry! 🤦 I was running the code in the wrong directory (`inline_fragment`, not `inline_fragment_bug`). My bad!

mat-sop:
I created pr with fix, where I included similar [test case](https://github.com/mirumee/ariadne-codegen/pull/151/files#diff-c0218ca9438d86f98f7a2bbd73645ec30fbcf36aec62d0de13bae1a75d5f2788R55).

Issue #144, opened 2023-05-15
Private fields are not added to classes


bombsimon:
Thanks for this awesome project!

According to the [Pydantic docs](https://docs.pydantic.dev/latest/usage/models/#automatically-excluded-attributes), private fields are automatically excluded from the object.

```python
class MyModel(BaseModel):
    public: str
    _private: str = Field(alias="private")


def main():
    d = {"public": "public", "private": "private"}
    m = MyModel.parse_obj(d) # Same with `MyModel(**d)` or `MyModel(public=..., _pvirate=...)`

    print(m1.public)
    print(m1._private)
```

Will print
```sh
public
default=PydanticUndefined alias='private' alias_priority=2 extra={}
```

Renaming the field to `private` (or anything dropping the leading `_`) makes this work as expected.

This is a general issue for any type that starts with an underscore but mostly interesting for me regarding `__typename` since that's a common practice for GraphQL. I saw you mentioned it in [this](https://github.com/mirumee/ariadne/discussions/605#discussioncomment-898336) discussion but also that the codegen has specific handling for this, f.ex. [here](https://github.com/mirumee/ariadne-codegen/blob/1edb477ae544d4fc825b28cc7c911f36eaeab773/ariadne_codegen/client_generators/result_types.py#L301-L303) and [here](https://github.com/mirumee/ariadne-codegen/blob/1edb477ae544d4fc825b28cc7c911f36eaeab773/ariadne_codegen/codegen.py#L284-L289). This means that accessing `__typename__` on any of the generated classes will only give me the `PydanticUndefined` value.

What's the intended way to access `__typename`? Is this a bug? How would I work with other fields starting with an underscore? For the GraphQL use case I don't see a reason to generate excluded fields, should they be renamed by default?

mat-sop:
Hey, thanks for pointing out this problem. I wasn't aware of this. In my opinion we can handle these fields by stripping leading underscores and keeping the original name as an alias. In addition let's change `__typename__` to something else, like `typename__`. I see that currently `convert_to_snake_case` option already handles some of these cases but not all of them.

bombsimon:
I actually noticed that for single prefixed underscore fields, this is resolved by using `convert_to_snake_case` which I think is enabled by default: https://github.com/mirumee/ariadne-codegen/blob/f61aedd6525744922b2e565a41d783fbf21c37b5/ariadne_codegen/utils.py#L32

So in a query like:

```grapql
query TestQuery {
    __typename
[...]

PR #148, opened 2023-05-17
fix: Don't prefix `typename` with underscore

bombsimon:
Rename `__typename__` so it doesn't start with an underscore. This is because fields with such prefix are marked as private in pydantic.

Ref #144

### DIFF
--- a/ariadne_codegen/client_generators/arguments.py
+++ b/ariadne_codegen/client_generators/arguments.py
@@ -63,6 +63,7 @@ class ArgumentsGenerator:
                 convert_to_snake_case=self.convert_to_snake_case,
                 plugin_manager=self.plugin_manager,
                 node=variable_definition,
+                trim_leading_underscore=True,
             )
             annotation, used_custom_scalar = self._parse_type_node(
                 variable_definition.type
--- a/ariadne_codegen/client_generators/arguments.py
+++ b/ariadne_codegen/client_generators/arguments.py
@@ -63,7 +63,6 @@ class ArgumentsGenerator:
                 convert_to_snake_case=self.convert_to_snake_case,
                 plugin_manager=self.plugin_manager,
                 node=variable_definition,
-                trim_leading_underscore=True,
             )
             annotation, used_custom_scalar = self._parse_type_node(
                 variable_definition.type
--- a/ariadne_codegen/utils.py
+++ b/ariadne_codegen/utils.py
@@ -92,7 +92,7 @@ def process_name(
     if iskeyword(processed_name):
         processed_name += "_"
     if trim_leading_underscore:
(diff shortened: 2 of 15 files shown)


ariadne_codegen/client_generators/arguments.py
```diff
@@ -63,6 +63,7 @@ def generate(
                 convert_to_snake_case=self.convert_to_snake_case,
                 plugin_manager=self.plugin_manager,
                 node=variable_definition,
+                trim_leading_underscore=True,
```
mat-sop, line 66: Imo we shouldn't trim underscores from arguments names, they are not part of pydantic's models so there is no harm in leaving them.

bombsimon: I'll of course change if you want to but I thought about it and this felt like the proper way if the type is changed in a class.

So f.ex. for this, I think it could've made sense to keep it:

```graphql
type Mutation {
  add_str(_arg: String!): String!
}

mutation TestStr($_arg: String!) {
  add_str(_arg: $_arg)
}
```

This would've generate this instead of `arg: str`:
```python
async def test_str(self, _arg: str) -> TestStr
```

However, if the return type would be a custom type like this:
```graphql
type T {
  _a: String!
}

input TInput {
  _a: String!
}

type Mutation {
  add_t(_t: TInput!): T!
}

mutation TestT($_t: TInput!) {
  add_t(_t: $_t) {
    _a
  }
}
```

We now generate 
```python
async def test_t(self, _t: TInput) -> TestT
```

And if we unwrap `TestT` we would have 
```python
class TestT(BaseModel):
    add_t: "TestTAddT"


class TestTAddT(BaseModel):
    a: str = Field(alias="_a")
```

Where `_a` that holds the string is now replaced.

I know they're different types (inputs vs. return types) but I think a consistent behaviour of GraphQL fields with leading underscores would make sense.

bombsimon: Actually after reading this, I think you're right, they're different things so no need to change this for non pydantic stuff! I'll change this!

mat-sop (approved):

ariadne_codegen/utils.py
```diff
@@ -90,6 +91,8 @@ def process_name(
         processed_name = str_to_snake_case(processed_name)
     if iskeyword(processed_name):
         processed_name += "_"
+    if trim_leading_underscore:
+        processed_name = re.sub("^_*", "", processed_name)
```
rafalp, line 95:
```suggestion
        processed_name = processed_name.lstrip("_")
```

rafalp (changes requested):

rafalp (approved):

PR #158, opened 2023-05-25
Set type hints for parse and serialize to `typing.Any`

bombsimon:
It's possible to use scalars for other types than `str`, f.ex. raw JSON with objects and/or lists or floats.

Given a scalar type that's not a `pydantic` class we might want to return a JSON serializable type like a `dict` or `list` from the serialize function.

Similarly we might want to accept such type when parsing it to our custom type.

---

We use custom scalars for raw JSON in some cases. Maybe this is a weird example but to demonstrate how this would be used.

```graphql
# schema.graphql
scalar MyScalar

type Mutation {
  create(input: MyScalar!): String!
}

# query.graphql
mutation TestMutation($input: MyScalar!) {
  create(input: $input)
}
```

```python
@dataclass
class SnowmanKeys:
    keys: list[str]
    value: str = "⛄️"


MyScalar = SnowmanKeys


def serialize_my_scalar(value: MyScalar):
    return {k: value.value for k in value.keys}
```

```toml
[tool.ariadne-codegen.scalars.MyScalar]
type = "scalars.MyScalar"
serialize = "scalars.serialize_my_scalar"
```

With this setup, given the following code

```python
await Client().test_mutation(input=SnowmanKeys(keys=["foo", "bar"]))
```

The generated GraphQL request would be

```graphql
{
    'query': 'mutation TestMutation($input: MyScalar!) { create(input: $input) }',
    'variables': {
        'input': { 'foo': '⛄️', 'bar': '⛄️' }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Removed `mixin` directive from operation string sent to server.
 - Fixed `ShorterResultsPlugin` that generated faulty code for discriminated unions.
 - Changed generator to ignore unused fragments which should be unpacked in queries.
+- Changed type hints for parse and serialize methods of scalars to `typing.Any`.
 
 
 ## 0.6.0 (2023-04-18)
--- a/ariadne_codegen/client_generators/dependencies/scalars.py
+++ b/ariadne_codegen/client_generators/dependencies/scalars.py
@@ -1,4 +1,4 @@
 from typing import Any, Callable, Dict
 
-SCALARS_PARSE_FUNCTIONS: Dict[Any, Callable[[str], Any]] = {}
-SCALARS_SERIALIZE_FUNCTIONS: Dict[Any, Callable[[Any], str]] = {}
+SCALARS_PARSE_FUNCTIONS: Dict[Any, Callable[[Any], Any]] = {}
+SCALARS_SERIALIZE_FUNCTIONS: Dict[Any, Callable[[Any], Any]] = {}
--- a/ariadne_codegen/client_generators/scalars.py
+++ b/ariadne_codegen/client_generators/scalars.py
@@ -120,14 +120,14 @@ class ScalarsDefinitionsGenerator:
                         name=SCALARS_PARSE_DICT_NAME,
                         dict_=self._parse_dict,
                         callable_annotation=generate_tuple(
-                            [generate_list([generate_name("str")]), generate_name(ANY)]
+                            [generate_list([generate_name(ANY)]), generate_name(ANY)]
                         ),
                     ),
                     self._generate_dict_assignment(
                         name=SCALARS_SERIALIZE_DICT_NAME,
                         dict_=self._serialize_dict,
                         callable_annotation=generate_tuple(
-                            [generate_list([generate_name(ANY)]), generate_name("str")]
+                            [generate_list([generate_name(ANY)]), generate_name(ANY)]
                         ),
                     ),
                 ],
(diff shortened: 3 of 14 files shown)


mat-sop (commented):
Do I understand correctly that your example works right now, but `mypy` produces error for  `serialize_my_scalar` in `scalars.py`?

Also this should be changed [here](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/client_generators/dependencies/scalars.py) as well.

bombsimon:
> Do I understand correctly that your example works right now, but `mypy` produces error for `serialize_my_scalar` in `scalars.py`?

Yes and now! It's true that it works now but in fact not even `mypy` reports this, I think it's because the method call is resolved dynamically so `mypy` doesn't know the caller from the generated code.

The reason I wanted to change this was purely because I read the code as some form of documentation because it took me quote some time to figure out when these methods was call. It's still not fully clear to me if and how `dataclasses` and `pydantic` classes will be automatically (de)serialized without passing these methods or not so right now I'm basically using something similar to my example. Having the type hints set correct would've avoid confusion for me and also be more representative of reality.

> Also this should be changed [here](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/client_generators/dependencies/scalars.py) as well.

Fixed!

mat-sop:
> The reason I wanted to change this was purely because I read the code as some form of documentation because it took me quote some time to figure out when these methods was call.

These methods are called in `base_model.py` - file attached to every generated client.


> It's still not fully clear to me if and how `dataclasses` and `pydantic` classes will be automatically (de)serialized without passing these methods or not so right now I'm basically using something similar to my example.

The idea was like this:
1. If type is a bultin (eg. `str`, `int`...) then you don't need to provide `serialize`/`parse` methods
2. Similar to 1. if pydantic can handle type then again you don't need to provide `serialize`/`parse` methods (eg. `dataclass`, `datetime.datetime` etc.). You still can if you want to customize how type is treated, like in your example, from your method you will get `{"key1": "⛄️", "key2": "⛄️"}` instead of default `{"keys": ["key1", "key2"], "value': "⛄️"}`.
3. If type is not handled by pydantic (eg. `class MyCustomType:`) then you need to provide `parse` and/or `serialize`.


While I was playing with your example a little bit, I found some inconsistency in implementation of 2. case, where we don't provide `serialize`, looks like right now we only support these "pydantic handled" types in results but not in inputs. I will address that in separate issue

mat-sop (approved):

bombsimon:
@mat-sop Thanks for the explanation!

I'm starting to ponder if doing what I do in the example above is more abusing the serialize method but it's really convenient to be able to go straight from what the API serves to what I want to work with internally. This also seems to be what the example does with going from a `-` separated string to an internal class representation.

One actual use case that I have though is that I have an API that requires a field prefixed with `__`, let's call it `__field`. If I create a dataclass like this:

```python
@dataclass
class MyScalar:
    __field: str
    another_field: str
```

Without a custom serialize method the generated request containing a `MyScalar` type would only contain `another_field` in the serialized output. So even though a simple class or dict can represent the type the default parsing and serialization will throw away fields I need. I'm not aware of a better way to solve this but maybe there's a better solution?

Either way I think since a scalar in GraphQL context isn't necessarily a string having the methods type hinted as `typing.Any` better represents custom scalars. Please let me know if you don't agree!

rafalp:
Please rebase on main branch

rafalp (approved):

PR #160, opened 2023-05-26
Support single underscore as a field name

aakog:
Without this fix, codegen fails with a parse error if the schema contains any fields named just `_`.

### DIFF
--- a/ariadne_codegen/utils.py
+++ b/ariadne_codegen/utils.py
@@ -93,6 +93,8 @@ def process_name(
         processed_name += "_"
     if trim_leading_underscore:
         processed_name = processed_name.lstrip("_")
+        if not processed_name:
+            processed_name = "underscore_"
     if plugin_manager:
         processed_name = plugin_manager.process_name(processed_name, node=node)
     return processed_name
--- a/tests/client_generators/input_types_generator/test_names.py
+++ b/tests/client_generators/input_types_generator/test_names.py
@@ -153,6 +153,7 @@ def test_generate_returns_module_with_valid_field_names():
         _foo: String!
         _Bar: String!
         ____baz_: String!
+        _: String!
     }
     """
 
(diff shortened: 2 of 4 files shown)


ariadne_codegen/utils.py
```diff
@@ -93,6 +93,8 @@ def process_name(
         processed_name += "_"
     if trim_leading_underscore:
         processed_name = processed_name.lstrip("_")
+        if not processed_name:
+            processed_name = "underscore_"
```
mat-sop, line 97: Hey, why `underscore_`? What when field with this name already exists? Tbh. In this case I would prefer to improve exception message and leave handling of it to plugins or using alias in query (eg. `underscore_: _`).
Btw. why schema has field named `_`?

aakog: > why underscore_?

`underscore_` is an arbitrary name. There may be a better name. I'd also be ok omitting the field.

> What when field with this name already exists?

`process_name()` already has issues with name collisions. This PR doesn't make it any better or worse.

> improve exception message and leave handling of it to plugins

I think this would work if it is acceptable for the tool to fail on valid graphql schemas with the default config. This kind of field worked in release 0.6.

> using alias in query

I'm ok with not supporting it in quieres. I only added query support for consistency. My main concern is if this field is in the schema. Currently if this field is in the schema, then `adriane-codegen` cannot be used at all for that API.

> why schema has field named `_`

My expectation as a user is that the codegen should work on any valid graphql name. Names like `in` and `_foo` are already supported, but `_` is not.

In the schema that I am working with `_` is a dummy field in a type that otherwise wouldn't have any fields. Graphql requires that all types have at least one field.

rafalp: > process_name() already has issues with name collisions. This PR doesn't make it any better or worse.

Does it really? From my understanding it will only create a collision if schema or queries are designed in a way that would cause those collisions in first place, eg. it has both `userName` and `user_name` fields next to each other. And as @bombsimon mentioned, we provide an escape hatch there through plugin system.

I would like to know more about your usecase. Is this a noop field in `Query` type that's never called?

I think we should have dedicated path before `if trim_leading_underscore:` that would delegate directly to plugin system and fallback to underscore otherwise:

```python
if set(name) == {"_"}:
    return plugin_manager.process_name(name, node=node) or "underscore_only_field_"
```

Default underscore name should be something obtuse like how React does `dangerouslySetInnerHTML`

PR #162, opened 2023-05-29
Serialize variables using pydantic_encoder

mat-sop:
This pr:
- adds [pytest-httpxs](https://github.com/Colin-b/pytest_httpx) as dev dependency
- changes base clients to use pydantic's [pydantic_encoder](https://docs.pydantic.dev/latest/usage/dataclasses/#json-dumping) for serialization of sent payload, without this change we handle some types for result fields but not for arguments, eg. `datetime.datetime` when someone provides it as scalar type (without custom `parse`/`serialize`), then result field with type `datetime.datetime` works fine, but argument with the same type causes exception

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Added `ShorterResultsPlugin` to standard plugins.
 - Fixed handling of inline fragments inside other fragments.
 - Changed generated unions to use pydantic's discriminated unions feature.
-- Used pydantic's `pydantic_encoder` to serialize `variables` dict.
+- Replaced HTTPX's `json=` serializer for query payloads with pydantic's `pydantic_encoder`.
 
 
 ## 0.6.0 (2023-04-18)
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Added `ShorterResultsPlugin` to standard plugins.
 - Fixed handling of inline fragments inside other fragments.
 - Changed generated unions to use pydantic's discriminated unions feature.
+- Used pydantic's `pydantic_encoder` to serialize `variables` dict.
```
rafalp, line 12:
```suggestion
- Replaced HTTPX's `json=` serializer for query payloads with pydantic's `pydantic_encoder`.
```

rafalp (approved):

mat-sop:
> This change has dropped the Content-Type header from the request (passing `content` instead of `json` to httpx). Was this intentional?

You're right, this wasn't my intention. I created issue for that #168

PR #181, opened 2023-07-03
Import fragment types from `.fragments`

### LINKED ISSUE #178: Incorrect import with top level fragment with ShorterResultsPlugin
Let's take example schema and query:
```gql
type Query {
  hello: TypeA!
}

type TypeA {
  valueB: TypeB!
}

type TypeB {
  id: ID!
}
```
```gql
query testQuery {
  ...fragmentHello
}

fragment fragmentHello on Query {
  hello {
    valueB {
      id
    }
  }
}
```

From these we generate `test_query.py`
```py
class TestQuery(FragmentHello):
    pass
```
and `fragments.py`
```py
class FragmentHello(BaseModel):
    hello: "FragmentHelloHello"


class FragmentHelloHello(BaseModel):
    value_b: "FragmentHelloHelloValueB" = Field(alias="valueB")
```

Without `ShorterResultsPlugin` we generate client which looks like this:
```py
from .test_query import TestQuery
...
    async def test_query(self) -> TestQuery:
        ...
        return TestQuery.parse_obj(data)
```

With  `ShorterResultsPlugin`:
```py
from .test_query import FragmentHelloHello, TestQuery
...
    async def test_query(self) -> FragmentHelloHello:
        ...
        return TestQuery.parse_obj(data).hello
```

Problem with client generated with `ShorterResultsPlugin` is that it imports `FragmentHelloHello` from `test_query.py`, but instead it should be imported from `fragments.py`.

bombsimon:
Resolves #178

This will use `generate_fragments_module` to find all classes that's generated in `fragments.py` and add them to `imported_types`. This will result in the imports in `client.py` to be imported from `.fragments` instead of the file where the original result class was found.


ariadne_codegen/contrib/shorter_results.py
```diff
@@ -99,6 +104,19 @@ def generate_result_class(
...
+    def generate_fragments_module(
+        self,
+        module: ast.Module,
+        fragments_definitions: Dict[str, FragmentDefinitionNode],
+    ) -> ast.Module:
+        """Store a map of all fragment classes and their AST."""
+        for fragment_class in [
+            x.name for x in module.body if isinstance(x, ast.ClassDef)
+        ]:
+            self.imported_types[fragment_class] = ".fragments"
```
mat-sop, line 116: `"fragments"` is only a default value, we shouldn't hardcode it. A dictionary with settings is passed to every `Plugin`, so you can get name like this:
```py
fragments_module_name = self.config_dict.get("tool", {}).get("ariadne-codegen", {}).get("fragments_module_name", "fragments")
```

mat-sop (changes requested):

ariadne_codegen/contrib/shorter_results.py
bombsimon: How should I approach the relative import path? When looking at this in test and my example I don't have a `fragments_module_name`. If that is present, would it be prefixed with a dot? And should I default to `.fragments` instead?

mat-sop: It's only a filename of a file that we generate, similar to `enums_module_name`, so it's safe to assume it will always be in generated package's directory. Either you can prefix it with `.` or change the [logic](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/contrib/shorter_results.py#L155) to set the [level](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/codegen.py#L28) of generated import to `1` when importing a fragment.

bombsimon: Cool, thanks! I opted to change the logic since we already have the dot/level set for the regular imports. Added a custom name for the test to ensure it gets picked up properly as well.

mat-sop (approved):

Issue #165, opened 2023-05-31
Feature request: set `default=None` on optional input type fields


acarapetis:
Consider the input type

```graphql
input User {
    username: String!
    fullName: String
}
```

for which `ariadne-codegen` generates an input model looking something like:

```python
class User(BaseModel):
    username: str
    full_name: Optional[str] = Field(alias="fullName")
```

If we try to instantiate this model using only a username, pyright complains:

```python
u = User(username="jen") # Argument missing for parameter "fullName"
```

Note that this code does work fine at runtime - unfortunately pydantic uses a slightly different convention to standard python dataclasses and defaults optional fields to None automatically, but pyright treats it as a dataclass-like type and thus expects all fields without defaults to be passed explicitly.

When I'm writing my own pydantic schemas, I always add an explicit `= None` to optional fields to avoid these false positives - would there be any issue with modifying ariadne-codegen to include this when appropriate?

mat-sop:
Hey, I looked into this, and it looks like this change won't cause any issues. I was concerned about behaviour of `exclude_unset` used in our base clients, but I did small [test](https://gist.github.com/mat-sop/65c6456fffb5bdfb5b67c308e53c18f3) and adding `= None` didn't make a difference.

Tbh I think that this is something that should be handled by pyright or pydantic, as we don't generate "unusual" models. If it's not possible to solve there, I would suggest creating a plugin to codegen. You can use [`generate_input_field`](https://github.com/mirumee/ariadne-codegen/blob/main/PLUGINS.md#generate_input_field) hook to easily alter default values of input fields, eg. 

 ```py
 import ast

from ariadne_codegen import Plugin
from graphql import GraphQLInputField


class DefaultValuesPlugin(Plugin):
    def generate_input_field(
        self,
        field_implementation: ast.AnnAssign,
        input_field: GraphQLInputField,
        field_name: str,
    ) -> ast.AnnAssign:
        if not is_optional(field_implementation):
            return field_implementation
        
        # Cases to handle:
        # 1. there is no field_implementation.value
        # 2. field_implementation.value represents Field(...)
        #   2.1. there is `default=` keyword in Field(...)
        #   2.2. Field(...) doesn't have `default` keyword 
        # 3. default value is already assigned

def is_optional(field_implementation: ast.AnnAssign) -> bool:
    ...

 ```

acarapetis:
[...]

PR #186, opened 2023-07-14
Pydantic v2

### LINKED ISSUE #183: Feature request: Pydantic V2
Pydantic has officially released Pydantic v2, it would be great if this code generator can emit Pydantic V2 code. 

https://github.com/pydantic/pydantic/releases/tag/v2.0.2

Duplicates #130

mat-sop:
This pr updates version of pydantic to `~2.0`. 

Changes in generated clients and models:
- `parse_obj` -> `model_validate`
- `.dict()` -> `.model_dump()`
- Instead of `Config` class `BaseModel` uses `model_config = ...` attribute.
- Refactored validator which handles parsing of custom scalars.
- `Upload` is no longer dataclass, in v2 pydantic's `.model_dump()` changes dataclasses to dicts which is harder for client to detect files.
- `update_forward_refs` -> `model_rebuild`
- `pydantic.json.pydantic_encoder` -> `pydantic_core.to_jsonable_python`
- Optional input fields now have explicit default `None` value (if schema doesn't specify another) #165

I see some potential to improve our implementation of custom scalars using some pydantic's v2 features, but this pr is already huge so it aims to sustain the same logic.

resolves #183 #165


ariadne_codegen/client_generators/dependencies/base_model.py
```diff
@@ -19,15 +16,17 @@ def __bool__(self) -> bool:
 
 
 class BaseModel(PydanticBaseModel):
-    class Config:
-        allow_population_by_field_name = True
-        validate_assignment = True
-        arbitrary_types_allowed = True
+    model_config = ConfigDict(
+        populate_by_name=True, validate_assignment=True, arbitrary_types_allowed=True
```
rafalp, line 20:
```suggestion
        populate_by_name=True,
        validate_assignment=True,
        arbitrary_types_allowed=True,
```

rafalp (approved):

Issue #187, opened 2023-07-19
Add Python 3.8 support


gst:
Hi, I'm interrested in using the project but would need python3.8 support still.

But I've seen that the generated code has `dict` annotation. while python3.8 only supports Dict.

Would you be ok to still consider python3.8 support ?  I have already a patch that for now seems to make the work (quickly tested).

Thx.

gst:
hi @mat-sop 

yes I/we can use python>=3.9 (3.10 actually) to generate that python code still ok with python3.8, thanks effectively to the 1 place you found (had also changed it in my clone).

would you want me to make that patch/PR (though for 1 line I guess you could make it) ?

cheers,

gst:
@mat-sop made tentative here: https://github.com/mirumee/ariadne-codegen/pull/188

had to change a bit more than I was expecting. not sure all of them are actually **required** but so, yes, I used a search/replace at one moment ;)

all tests pass localy on my system at least.

PR #188, opened 2023-07-24
Add/keep python3.8 support (dict -> Dict) for generated code

gst:
Fix: https://github.com/mirumee/ariadne-codegen/issues/187

### DIFF
--- a/ariadne_codegen/client_generators/client.py
+++ b/ariadne_codegen/client_generators/client.py
@@ -263,7 +263,7 @@ class ClientGenerator:
         return generate_ann_assign(
             target=self._variables_dict_variable,
             annotation=generate_subscript(
-                generate_name("Dict"),
+                generate_name(DICT),
                 generate_tuple([generate_name("str"), generate_name("object")]),
             ),
             value=arguments_dict,
--- a/ariadne_codegen/client_generators/client.py
+++ b/ariadne_codegen/client_generators/client.py
@@ -32,6 +32,7 @@ from .arguments import ArgumentsGenerator
 from .constants import (
     ANY,
     ASYNC_ITERATOR,
+    DICT,
     LIST,
     MODEL_VALIDATE_METHOD,
     OPTIONAL,
@@ -65,7 +66,7 @@ class ClientGenerator:
         self._imports: List[ast.ImportFrom] = []
         self._add_import(
             generate_import_from(
-                [OPTIONAL, LIST, ANY, UNION, ASYNC_ITERATOR], TYPING_MODULE
+                [OPTIONAL, LIST, DICT, ANY, UNION, ASYNC_ITERATOR], TYPING_MODULE
             )
         )
         self._add_import(base_client_import)
@@ -262,7 +263,7 @@ class ClientGenerator:
         return generate_ann_assign(
             target=self._variables_dict_variable,
             annotation=generate_subscript(
-                generate_name("dict"),
+                generate_name("Dict"),
                 generate_tuple([generate_name("str"), generate_name("object")]),
             ),
             value=arguments_dict,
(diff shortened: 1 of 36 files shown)


ariadne_codegen/client_generators/client.py
```diff
@@ -262,7 +263,7 @@ def _generate_variables_assign(
         return generate_ann_assign(
             target=self._variables_dict_variable,
             annotation=generate_subscript(
-                generate_name("dict"),
+                generate_name("Dict"),
```
mat-sop, line 266: We should use the constant as well
```suggestion
                generate_name(DICT),
```

tests/client_generators/test_client_generator.py
```diff
@@ -295,7 +297,7 @@ def test_add_method_generates_correct_async_method_body(
         ast.AnnAssign(
             target=ast.Name(id="variables"),
             annotation=ast.Subscript(
-                value=ast.Name(id="dict"),
+                value=ast.Name(id="Dict"),
```
mat-sop, line 300:
```suggestion
                value=ast.Name(id=DICT),
```

tests/client_generators/test_client_generator.py
```diff
@@ -449,7 +451,7 @@ def test_add_method_generates_correct_method_body(
...
```
mat-sop, line 454:
```suggestion
                value=ast.Name(id=DICT),
```

tests/client_generators/test_client_generator.py
```diff
@@ -545,7 +547,7 @@ def test_add_method_generates_async_generator_for_subscription_definition(
             ast.AnnAssign(
                 target=ast.Name(id="variables"),
                 annotation=ast.Subscript(
-                    value=ast.Name(id="dict"),
+                    value=ast.Name(id="Dict"),
```
mat-sop, line 550:
```suggestion
                    value=ast.Name(id=DICT),
```

ariadne_codegen/graphql_schema_generators/utils.py
```diff
@@ -13,11 +13,19 @@
...
+        return "Tuple"
+    return obj.__class__.__name__
+
+
 def get_named_type(type_: GraphQLNamedType, type_map_name: str) -> ast.Call:
     return generate_call(
         func=generate_name("cast"),
         args=[
-            generate_name(type_.__class__.__name__),
+            generate_name(_keep_dict_tuple_capitalized(type_)),
```
mat-sop, line 28: Hmm, I don't see a point in this change. `type_` has the annotation `GraphQLNamedType`, so it shouldn't get `dict` or `tuple`. Am I missing something? Do you see a case where it's necessary?

mat-sop (changes requested):

gst:
hmm @mat-sop 

ideally this would not deserve a python3.8 build/CI (run) test configuration too ?

mat-sop (approved):

mat-sop:
> hmm @mat-sop
> 
> ideally this would not deserve a python3.8 build/CI (run) test configuration too ?

Ideally yes, but most of our tests check if the generated package looks like expected and we can't run those tests on `3.8`
To include `3.8` in our ci we would need to:
1. Use separate python versions for generation and client tests.
2. Create tests that check only the behavior of the generated client.

`1.` seems relatively easy to do, but `2.` definitely requires more work, so it may be something to consider in the future. Of course, if you want to create those tests, contributions are very welcome 😄

gst:
> Use separate python versions for generation and client tests.

yes obviously :)

> Create tests that check only the behavior of the generated client.

that would be out of the scope of this PR ;)

but you could have simple checks, kind of : 

1) import every generated sub-module, must succeed, obviously.
2) egrep "[^\w](dict|tuple)\[" **/*.py ; for not top-level annotations.

eventually.

rafalp (approved):
LGTM. We'll need to remember to mention this change in changelog after this is merged.

PR #196, opened 2023-08-21
Handle field names reserved by pydantic

### LINKED ISSUE #191: NameError: Field name "schema" shadows a BaseModel attribute; use a different field name with "alias='schema'".
Schema that use a field called "schema" fail the pydantic post generation check. See https://pythonfix.com/error/pydantic-field-name-schema-shadows-basemodel-attribute/ and https://stackoverflow.com/questions/73828037/pydantic-field-does-not-take-value it would be nice if the generator worked around this by automatically renaming the field to "schema_" and adding = Field(..., alias='schema')

mat-sop:
This pr handles input and result field names which were colliding with names reserved by pydantic's `BaseModel` methods and attributes. The same as python keywords, we add `_` as a suffix. Full list of reserved names:
```python
In [2]: [name for name in dir(BaseModel) if not name.startswith("_")]
Out[2]: 
['construct',
 'copy',
 'dict',
 'from_orm',
 'json',
 'model_computed_fields',
 'model_config',
 'model_construct',
 'model_copy',
 'model_dump',
 'model_dump_json',
 'model_extra',
 'model_fields',
 'model_fields_set',
 'model_json_schema',
 'model_parametrized_name',
 'model_post_init',
 'model_rebuild',
 'model_validate',
 'model_validate_json',
 'parse_file',
 'parse_obj',
 'parse_raw',
 'schema',
 'schema_json',
 'update_forward_refs',
 'validate']
```

resolves #191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Changed generated client and models to use pydantic v2.
 - Changed custom scalars implementation to utilize pydantic's `BeforeValidator` and `PlainSerializer`. Added `scalars_module_name` option. Replaced `generate_scalars_parse_dict` and `generate_scalars_serialize_dict` with `generate_scalar_annotation` and `generate_scalar_imports` plugin hooks.
 - Fixed generating default values of input types from remote schemas.
+- Changed generating of input and result field names to add `_` to names reserved by pydantic.
 
 
 ## 0.7.1 (2023-06-06)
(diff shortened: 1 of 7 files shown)


rafalp (approved):

josuemontano:
Thanks for this update! I'm curious if there's a possibility of backporting this to v0.7.x for the benefit of users who are still on pydantic 1. Thanks for the work!

PR #205, opened 2023-09-05
Fix generating result types with nullable directives

### LINKED ISSUE #204: Add full support for skip/include directives
GQL spec defines [skip](https://spec.graphql.org/June2018/#sec--skip) and [include](https://spec.graphql.org/June2018/#sec--include) directives. They are not fully supported cause generated pydantic types cannot parse server-side responses (e.g. skipped part of the query is missing and pydantic raises validation error).

**Example reproduction steps:**
Using https://beta.pokeapi.co/graphql/console/:

```graphql
query getPokemonByName($name: String!, $includeAbilities: Boolean!) {

  pokemon_v2_pokemon(where: {
    name: {
      _eq: $name
    }
  }) {
    id
    name
    pokemon_v2_pokemonabilities @include(if: $includeAbilities) {
      pokemon_v2_ability {
        id
        name
      }
    }
  }

}
```

```
>>> result = client.get_pokemon_by_name('pikachu', include_abilities=False)

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jbacic/Documents/Private/ariadne-codegen/app/graphql_client/client.py", line 40, in get_pokemon_by_name
    return GetPokemonByName.model_validate(data)
  File "/Users/jbacic/Documents/Private/ariadne-codegen/.venv/lib/python3.10/site-packages/pydantic/main.py", line 504, in model_validate
    return cls.__pydantic_validator__.validate_python(
pydantic_core._pydantic_core.ValidationError: 1 validation error for GetPokemonByName
pokemon_v2_pokemon.0.pokemon_v2_pokemonabilities
  Field required [type=missing, input_value={'id': 25, 'name': 'pikachu'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.3/v/missing
```

The following server-side response:
```json
{
  "data": {
    "pokemon_v2_pokemon": [
      {
        "id": 25,
        "name": "pikachu"
      }
    ]
  }
}
```
cannot be parsed by:
```python
class GetPokemonByName(BaseModel):
    pokemon_v2_pokemon: List["GetPokemonByNamePokemonV2Pokemon"]


class GetPokemonByNamePokemonV2Pokemon(BaseModel):
    id: int
    name: str
    pokemon_v2_pokemonabilities: Optional[
        List["GetPokemonByNamePokemonV2PokemonPokemonV2Pokemonabilities"]
    ]
```
because it doesn't contain `pokemon_v2_pokemonabilities` key.

jakub-bacic:
Assign `None` as default value for optional fields created due to nullable directives. It's necessary to do so explicitly cause otherwise `pydantic` will complain about missing field (it's a regression introduced due to migration to pydantic v2).

Resolves #204.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fixed generating operation string for nested inline fragments.
 - Removed scalars module. Changed generated models and client to use annotated types for custom scalars. Removed `scalars_module_name` option. Removed `generate_scalars_module`, `generate_scalars_cod`, `generate_scalar_annotation` and `generate_scalar_imports` plugin hooks.
 - Removed pydantic warnings for fields with `model_` prefix.
+- Fixed generating result types with nullable directives.
 
 
 ## 0.8.0 (2023-08-22)
--- a/ariadne_codegen/client_generators/result_fields.py
+++ b/ariadne_codegen/client_generators/result_fields.py
@@ -61,10 +61,11 @@ def parse_operation_field(
     typename_values: Optional[List[str]] = None,
     custom_scalars: Optional[Dict[str, ScalarData]] = None,
     fragments_definitions: Optional[Dict[str, FragmentDefinitionNode]] = None,
-) -> Tuple[Annotation, List[FieldNames]]:
+) -> Tuple[Annotation, Optional[ast.expr], List[FieldNames]]:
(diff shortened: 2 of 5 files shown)


mat-sop (approved):
Could you also add an entry to the changelog?

rafalp (approved):

PR #219, opened 2023-09-20
Added missing dependency to pyproject.toml

danieledwardgeorgehitchcock:
When running mypy against the generated client in a fresh environment, the following warnings are raised:

```
graphql_client\async_base_client.py:19: error: Cannot find implementation or library stub for module named "websockets.client"  [import]
graphql_client\async_base_client.py:19: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
graphql_client\async_base_client.py:20: error: Cannot find implementation or library stub for module named "websockets.typing"  [import]
```

I have added the missing dependency to the `pyproject.toml`, have tested locally and this error then disappears when included.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
   "black",
   "isort",
   "autoflake",
+  "websockets~=11.0.3",
 ]
 
 [project.optional-dependencies]


mat-sop (changes requested):
We have `websockets~=11.0` as an optional [`subscriptions`](https://github.com/mirumee/ariadne-codegen/blob/main/pyproject.toml#L47) dependency. We should modify [`async_base_client.py`](https://github.com/mirumee/ariadne-codegen/blob/main/ariadne_codegen/client_generators/dependencies/async_base_client.py) to pass `mypy --strict` even without installed `websockets`.

rafalp:
Superseded by #246

PR #221, opened 2023-09-22
Stable comments

### LINKED ISSUE #215: First comment in auto-generated files causes conflicts in git
The first comment in auto-generated files, e.g.: `# Generated by ariadne-codegen on 2023-09-18 16:52`, causes problems when two people auto-generate the same file with different content. Changes that could easily be merged in git without the comment, result in conflicts when the comments are there. Setting `include_comments` to `false` in config is not ideal, as then there is no info for code reviewers etc. that the code in a given file was auto-generated. It should be possible to keep the comments, but without the date (anyway, the approximate date of the change can be found in git).

mat-sop:
This pr:
- changes `include_comments` option to accept `"stable"`, `"timestamp"` or `"none"`
- deprecates using boolean, `false` is mapped to `"none"` and `true` to `"timestamp"`
- changes default to `"stable"`
- adds `get_comment` plugin hook

resolves #215


PLUGINS.md
```diff
@@ -285,6 +285,15 @@ Hook executed on creating `GraphQLSchema` object from path or url provided in se
 To ensure all plugins have current version of schema, result of this hook is propagated to all plugins, updating their `schema` field. 
 
 
+### get_comment
+
+```py
+def get_comment(self, comment: str, code: str, source: Optional[str] = None) -> str:
```
rafalp, line 291:
```suggestion
def get_file_comment(self, comment: str, code: str, source: Optional[str] = None) -> str:
```

README.md
```diff
@@ -73,7 +73,7 @@ Optional settings:
 - `enums_module_name` (defaults to `"enums"`) - name of file with generated enums models
 - `input_types_module_name` (defaults to `"input_types"`) - name of file with generated input types models
 - `fragments_module_name` (defaults to `"fragments"`) - name of file with generated fragments models
-- `include_comments` (defaults to `true`) - a flag that specifies whether to include comments in generated files
+- `include_comments` (defaults to `"stable"`) - option which sets content of included comments, valid choices are: `"none"`, `"timestamp"`, `"stable"`
```
rafalp, line 76:
```suggestion
- `include_comments` (defaults to `"stable"`) - option which sets content of comments included at the top of every generated file. Valid choices are: `"none"` (no comments), `"timestamp"` (comment with generation timestamp), `"stable"` (comment contains a message that this is a generated file)
```

rafalp (changes requested):

PLUGINS.md
mat-sop: Done

rafalp (approved):

PR #223, opened 2023-09-25
Change `str_to_snake_case` to handle capitalized words

### LINKED ISSUE #213: All Upper case for snake case convert
If a query is named `GetBusinessID`, it will be converted to `get_business_i_d`. The better way would be `get_business_id`.

Instead of current implementation in `utils/str_to_snake_case`, I tried to use this regex:

```regex
(?<!^)([A-Z][a-z]|(?<=[a-z])[^a-z]|(?<=[A-Z])[0-9_])(?<!_)
```

But it would also convert from `arg1` to `arg_1`. Though it looks better, this brings lots of breaking in tests. I'm not sure this is a right way since users' code will need many modifications.

Is there any better ways to convert it?

mat-sop:
resolves #213


rafalp (approved):

rafalp:
This is a breaking change. We should mention in changelog that codegen will now prepend digits with underscore.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - Fixed generating results for nullable fields with nullable directives.
 - Changed `include_comments` option to accept enum value, changed default to `"stable"`, deprecated boolean support. Added `get_file_comment` plugin hook.
-- Changed `str_to_snake_case` utility to correctly handle capitalized words.
+- Changed `str_to_snake_case` utility to correctly handle capitalized words. This is a breaking change, digits are now preceded by an underscore.
```
rafalp, line 7:
```suggestion
- Changed `str_to_snake_case` utility to correctly handle capitalized words.
- Digits in Python names are now preceded by an underscore (breaking change).
```

PR #227, opened 2023-10-06
Add opentelemetry to base clients

### LINKED ISSUE #200: OpenTelemetry support
Generated client could support opt-in OpenTelemetry spans, to track number and time of made GraphQL calls by application using it.

mat-sop:
This pr adds opentelemetry to both base clients

resolves #200


rafalp:
I am wondering if this code should only be included if feature flag is enabled 🤔 But that would explode the complexity by a lot?

mat-sop:
> I am wondering if this code should only be included if feature flag is enabled 🤔 But that would explode the complexity by a lot?

Hmm, we could treat them as separate base clients, e.g. `BaseClient`, `AsyncBaseClient`, `BaseClientWithTelemetry`, `AsyncBaseClientWithTelemetry` and decide which one to use based on 2 options - `async_client` (already exists) and something like `include_telemetry`.

This wouldn't add much complexity as we already have the option to provide a fully custom base client, it's only a matter of providing a path to a different file.

rafalp (changes requested):
Let's keep base clients base, and introduce separate clients for telemetry.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Digits in Python names are now preceded by an underscore (breaking change).
 - Fixed parsing of unions and interfaces to always add `__typename` to generated result models.
 - Added escaping of enum values which are Python keywords by appending `_` to them.
+- Added additional base clients with opentelemetry support. Added `telemetry_client` config option.
```
rafalp, line 11:
```suggestion
- Added additional base clients supporting the Open Telemetry tracing. Added `opentelemetry_client` config option.
```

README.md
```diff
@@ -142,6 +143,20 @@ type = "Upload"
 ```
 
 
+### Opentelemetry
```
rafalp, line 146:
```suggestion
### Open Telemetry
```

README.md
```diff
@@ -76,6 +76,7 @@ Optional settings:
 - `include_comments` (defaults to `"stable"`) - option which sets content of comments included at the top of every generated file. Valid choices are: `"none"` (no comments), `"timestamp"` (comment with generation timestamp), `"stable"` (comment contains a message that this is a generated file)
 - `convert_to_snake_case` (defaults to `true`) - a flag that specifies whether to convert fields and arguments names to snake case
 - `async_client` (defaults to `true`) - default generated client is `async`, change this to option `false` to generate synchronous client instead
+- `telemetry_client` (defaults to `false`) - default base clients doesn't support opentelemetry, change this option to `true` to use base client with opentelemetry
```
rafalp, line 79:
```suggestion
- `opentelemetry_client` (defaults to `false`) - default base clients don't support any performance tracing. Change this option to `true` to use the base client with Open Telemetry support.
```

README.md
```diff
@@ -142,6 +143,20 @@ type = "Upload"
 ```
 
 
+### Opentelemetry
+
+When config option `telemetry_client` is set to `true` then default, included base clients support opt-in telemetry options. By default, it's disabled, but when the `opentelemetry-api` package is installed and the `tracer` argument is provided then the client will create spans with data about performed requests.
```
rafalp, line 148:
```suggestion
When config option `opentelemetry_client` is set to `true` then default, included base client is replaced with one that implements the opt-in Open Telemetry support. By default this support does nothing but when the `opentelemetry-api` package is installed and the `tracer` argument is provided then the client will create spans with data about performed requests.
```

.github/workflows/tests.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install wheel
-        pip install -e .[subscriptions,dev]
+        pip install -e .[subscriptions,telemetry,dev]
```
rafalp, line 28:
```suggestion
        pip install -e .[subscriptions,opentelemetry,dev]
```

rafalp (changes requested):
"Telemetry" is unspecific in world of APM solutions, so we should use `opentelemetry` instead of `telemetry` for variable/option/class names.

For example `BaseClientOpenTelemetry` would be more descriptive than `BaseClientWithTelemetry`.

I understand some names will java'squely long, but their meaning would save us confusion in future.

README.md
```diff
@@ -142,6 +143,20 @@ type = "Upload"
 ```
 
 
+### Open Telemetry
+
+When config option `opentelemetry_client` is set to `true` then default, included base client is replaced with one that implements the opt-in Open Telemetry support. By default this support does nothing but when the `opentelemetry-api` package is installed and the `tracer` argument is provided then the client will create spans with data about performed requests.
+
+Telemetry arguments handled by `BaseClientOpenTelemetry`:
```
rafalp, line 150:
```suggestion
Tracing arguments handled by `BaseClientOpenTelemetry`:
```

rafalp (approved):

PR #242, opened 2023-11-21
No reimports plugin

### LINKED ISSUE #233: Load of module and client extremely slow.
## Description

I have a fairly complex GraphQL repository running on Hasura, that might have hundreds if not thousands of types and definitions. Right now I implemented this library for some specific queries and use cases, and all of those types available in the server are not needed for the execution of my isolated use case.

The generated code, creates a `__init__.py` file where everything is imported on package load. This `__init__.py` file is easily 4k lines of code, generating thousands of imports. Loading the `client` module might easily take minutes and GBs of RAM due to this reason. So I went ahead and deleted everything from the `__init__.py` and left it empty, as the calls to the right types happen on a module by module basis. This reduced my loading time from minutes to miliseconds, and the application still works as expected.

Now the question: is there a reason to import absolutely everything on the `__init__.py`?  For big and complex schemas is an overkill, and increases the application loading time by minutes. Should this file be empty and do a module by module import? - if that is the case, is this not already happening in the current implementation?

Maybe I am missing something, and I greatly appreciate your time and consideration.

mat-sop:
This pr adds `NoReimportsPlugin` which removes the content of generated `__init__.py`.

resolves #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 - Removed `model_rebuild` calls for generated input, fragment and result models. 
+- Added `NoReimportsPlugin` that removes content of `__init__.py`.
 
 
 ## 0.10.0 (2023-11-15)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## UNRELEASED
 
 - Removed `model_rebuild` calls for generated input, fragment and result models. 
-- Added `NoReimportsPlugin` that removes content of `__init__.py`.
+- Added `NoReimportsPlugin` that makes the `__init__.py` of generated client package empty.
 
 
 ## 0.10.0 (2023-11-15)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 - Removed `model_rebuild` calls for generated input, fragment and result models. 
+- Added `NoReimportsPlugin` that removes content of `__init__.py`.
```
rafalp, line 6:
```suggestion
- Added `NoReimportsPlugin` that makes the `__init__.py` of generated client package empty.
```

README.md
```diff
@@ -103,6 +103,8 @@ plugins = ["ariadne_codegen.contrib.extract_operations.ExtractOperationsPlugin"]
 operations_module_name = "custom_operations_module_name"
 ```
 
+- [`ariadne_codegen.contrib.extract_operations.NoReimportsPlugin`](ariadne_codegen/contrib/no_reimports.py) - This plugin removes content of generated `__init__.py`.
```
rafalp, line 106:
```suggestion
- [`ariadne_codegen.contrib.extract_operations.NoReimportsPlugin`](ariadne_codegen/contrib/no_reimports.py) - This plugin removes content of generated `__init__.py`. This is useful in scenarios where generated plugins contain so many Pydantic models that client's eager initialization of entire package on first import is very slow.
```

rafalp (approved):

PR #245, opened 2023-12-01
Add operationName

### LINKED ISSUE #225: Feature request: Include operationName if available in all requests
Hi!

Currently, only `query` and `variables` (if available) are attached to the request body of any outgoing requests in the base client.

It would make sense to add `operationName` as part of the payload. I know it's part of the `query` when present, but for example [graphql-request attaches it to the body of the outgoing request](https://github.com/jasonkuhrt/graphql-request/blob/main/src/index.ts#L268) ([and here](https://github.com/jasonkuhrt/graphql-request/blob/main/src/index.ts#L589)). And it's good practice to have it easily accessible for e.g. query logs & server side debugging.

In my mind, the implementation would work similarly to how the `variables` is attached currently.

E.g. something like:
```        
if operation_name:
    payload["operationName"] = operation_name
```

Thanks again for this great project! ⭐

mat-sop:
This pr adds `operationName` key to the payload sent to a server. We already require all parsed operations to have a name. In generated methods, it is passed as string literal, e.g:
```python
response = await self.execute(query=query, operation_name="CreateUser", variables=variables, **kwargs)
```

resolves #225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Added `NoReimportsPlugin` that makes the `__init__.py` of generated client package empty.
 - Added `include_all_inputs` config flag to generate only inputs used in supplied operations.
 - Added `include_all_enums` config flag to generate only enums used in supplied operations.
-- Changed generated client's methods to add `operationName` to sent payload.
+- Added `operationName` to payload sent by generated client's methods.
 
 
 ## 0.10.0 (2023-11-15)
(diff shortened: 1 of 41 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added `NoReimportsPlugin` that makes the `__init__.py` of generated client package empty.
 - Added `include_all_inputs` config flag to generate only inputs used in supplied operations.
 - Added `include_all_enums` config flag to generate only enums used in supplied operations.
+- Changed generated client's methods to add `operationName` to sent payload.
```
rafalp, line 9:
```suggestion
- Added `operationName` to payload sent by generated client's methods.
```

rafalp (approved):

PR #246, opened 2023-12-04
Mypy strict without optional dependencies

mat-sop:
This pr fixes base clients to pass `mypy --strict` without optional dependencies installed.


mat-sop:
> Does this PR resolve #219?

Yes, it resolves #219

PR #256, opened 2023-12-21
[bug] Fix scalar data issue in plugin settings

slessans:
For some reason when loading the config during plugin execution, `ScalarData` has already been converted which causes this code to throw an error.  I added a check to take the `ScalarData` as is if it has already been converted.


mat-sop (changes requested):
Hey, it looks like the issue is that in this method we are directly modifying the argument `config_dict`. Instead of the proposed check, I would rather see a change that leaves the dict unmodified after the first call (for example by working on a copy)

slessans:
@mat-sop updated!

rafalp (approved):

rafalp:
Thank you!

PR #257, opened 2023-12-21
Fix enum name processing and add uppercase flag

slessans:
This PR fixes the following bugs in the processing of enum names:

1. the snake_case conversion function did not properly handle names with all uppercase eg `DESC_NULLS_LAST` was being converted into `LAST`
2. The enum generator was not calling `process_name` and so could not be hooked into the plugin customization pipeline
3. The enum generator did not respect the `convert_to_snake_case` option

This PR additionally adds the `convert_enums_names_to_upper_snake_case` option to convert enum members to upper snake case like:

```python
enum Foo:
  FOO_BAR
  BAZ_BOW
```

I think this should be a core option rather than a plugin since it is more idiomatic. It also avoid enum name conflicts. For instance I had an enum with `title` and `name` as members, which conflicted with the built-in enum methods of the same name. I set the flag to default False for backwards compat.

### DIFF
--- a/ariadne_codegen/client_generators/enums.py
+++ b/ariadne_codegen/client_generators/enums.py
@@ -12,15 +12,21 @@ from ..codegen import (
     generate_module,
 )
 from ..plugins.manager import PluginManager
+from ..utils import process_name
 from .constants import ENUM_CLASS, ENUM_MODULE
 
 
 class EnumsGenerator:
     def __init__(
-        self, schema: GraphQLSchema, plugin_manager: Optional[PluginManager] = None
+        self, schema: GraphQLSchema, 
+        convert_to_snake_case: bool = False, 
+        convert_to_upper_snake_case: bool = False, 
+        plugin_manager: Optional[PluginManager] = None
     ) -> None:
         self.schema = schema
         self.plugin_manager = plugin_manager
+        self.convert_to_snake_case = convert_to_snake_case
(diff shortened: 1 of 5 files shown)


mat-sop (commented):
Hey, it looks like a lot of different topics for 1 pr:

> 1. the snake_case conversion function did not properly handle names with all uppercase eg `DESC_NULLS_LAST` was being converted into `LAST`

This looks quite serious, but I'm not sure if we should mark this as a breaking change, or just a fix. 

> 2. The enum generator was not calling `process_name` and so could not be hooked into the plugin customization pipeline

You can use `generate_enum` hook, but I agree that using `process_name`  here can be a nice improvement.

> 3. The enum generator did not respect the `convert_to_snake_case` option
>
> This PR additionally adds the `convert_enums_names_to_upper_snake_case` option to convert enum members to upper snake case like:

The fact that `convert_to_snake_case` doesn't apply to enums definitely should be better documented. I'm not sure if we should implement it, same for `convert_enums_names_to_upper_snake_case`. In your implementation you can't have current behavior - snake case for arguments and result/input fields, not changed enums. Maybe it should be a plugin?

> For instance I had an enum with `title` and `name` as members, which conflicted with the built-in enum methods of the same name.

We should escape those by adding `_`, similar to how we already handle overlap with pydantic model's attributes and method.

@rafalp what do you think?

PR #258, opened 2023-12-22
`model_rebuild` calls for top level fragments

### LINKED ISSUE #253: [bug] Pydantic model definition errors for fragments with nesting
## Description

Pydantic types are incorrectly defined when fragments contain nested nodes.

Example:

```graphql
# works
fragment IssueFields on Issue {
    id
}

query getIssue($id: ID!) {
    node(id: $id) {
        ...IssueFields
    }
}

# works
query getIssue($id: ID!) {
    node(id: $id) {
        ...on Issue {
            id
            author {
                login
            }
        }
    }
}

# does not work
fragment IssueFields on Issue {
    id
    author {
        login
    }
}

query getIssue($id: ID!) {
    node(id: $id) {
        ...IssueFields
    }
}
```

There are two issues in the last case.

1. The generated `IssueFieldsAuthor` type includes a required `__typename` field that is not included in the generated query, causing a validation error.
2. The generated `IssueFields` type contains an `author` field of type `Optional["IssueFieldsAuthor"]` that somehow seems to be undefined at instantiation, yielding the following error: 
```
pydantic.errors.PydanticUserError: `GetIssue` is not fully defined; you should define `GetIssueNodeNode`, then call `GetIssue.model_rebuild()`
```

The first issue disappears when including `__typename` in the generated query (in the client `get_issue` method).

The second issue disappears when defining `IssueFieldsAuthor` before `IssueFields` in `fragments.py`.

## Reproduction

1. Define these queries

```graphql
# queries.graphql

fragment IssueFields on Issue {
    id
    author {
        login
    }
}

query getIssue($id: ID!) {
    node(id: $id) {
        ...IssueFields
    }
}
```

Use the [GitHub schema](https://docs.github.com/en/graphql/overview/public-schema).

Generate the types with `ariadne-codegen`.

Run the following:

```python
import asyncio

# use the path to the generated client here
from src.graphql.graphql_client import Client


async def run():
    access_token = "<YOUR TOKEN>"
    client = Client(
        url="https://api.github.com/graphql",
        headers={
            "Authorization": f"Bearer {access_token}",
            "Content-Type": "application/json",
            "Accept": "application/vnd.github.bane-preview+json",
        },
    )

    issue = await client.get_issue(id="I_kwDOIPDwls55yyZU")

    print(issue)


if __name__ == "__main__":
    asyncio.run(run())
```

## Environment

Python 3.11.6

mat-sop:
This pr restores `model_rebuild` calls for top-level fragment models. When a fragment is parsed to many models only 1 (with the same name as the fragment) will get `model_rebuild` call.

Implementation is complete, but I'm marking this as a draft because I think it's a good opportunity to add a test that checks if generated models can be imported and used with `.model_validate(...)`. Maybe as a part of [generated_client](https://github.com/mirumee/ariadne-codegen/tree/main/tests/main/generated_client) or new test that imports models from [expected_clients](https://github.com/mirumee/ariadne-codegen/tree/main/tests/main/clients)?

resolves #253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## UNRELEASED
+
+- Restored `model_rebuild` calls for top level fragment models.
+
+
 ## 0.11.0 (2023-12-05)
 
 - Removed `model_rebuild` calls for generated input, fragment and result models. 
--- a/EXAMPLE.md
+++ b/EXAMPLE.md
@@ -501,6 +501,10 @@ class BasicUser(BaseModel):
 class UserPersonalData(BaseModel):
     first_name: Optional[str] = Field(alias="firstName")
     last_name: Optional[str] = Field(alias="lastName")
+
+
+BasicUser.model_rebuild()
+UserPersonalData.model_rebuild()
 ```
 
 ### Init file
--- a/ariadne_codegen/client_generators/constants.py
+++ b/ariadne_codegen/client_generators/constants.py
@@ -56,6 +56,7 @@ DISCRIMINATOR_KEYWORD = "discriminator"
 MODEL_VALIDATE_METHOD = "model_validate"
 PLAIN_SERIALIZER = "PlainSerializer"
 BEFORE_VALIDATOR = "BeforeValidator"
+MODEL_REBUILD_METHOD = "model_rebuild"
 
 ENUM_MODULE = "enum"
 ENUM_CLASS = "Enum"
--- a/ariadne_codegen/client_generators/fragments.py
+++ b/ariadne_codegen/client_generators/fragments.py
@@ -3,9 +3,9 @@ from typing import Dict, List, Optional, Set, cast
 
 from graphql import FragmentDefinitionNode, GraphQLSchema
 
-from ..codegen import generate_module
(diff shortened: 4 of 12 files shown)


rafalp:
@bombsimon I'll be taking over this (and other) PRs from Matt who left the company. I'll need to catch up with the code and internals first, but this PR is a priority for me.

bombsimon:
@rafalp Oh, sad to hear you're losing a great colleague but congrats to Matt I hope! 🎈 And luckily for us using this the codebase is still in good hands with you!

Understandable that this will take some time, let me know if I can do anything to help with the process! And FYI I based my branch with issues on this branch and that works just fine so this solves the problem for me.

rafalp:
Hello @bombsimon. If you want to pick up this PR to add tests to it, please feel free to do so.

Apologies for taking this long to return to this.

bombsimon:
@rafalp I'll try to pick this up next week and I'll let you know if I won't make it so you're not waiting for me!

Don't mention it, all your effort is greatly apprecaited!

rafalp:
> Maybe a bit out of scope since this should be tested

Tests like this are useful, they codify assumptions/requirements on the dependency's behavior in tests suite.

rafalp:
Superseded by #267

PR #259, opened 2024-01-07
Add scalar to Operation plugin client test

alzex3:
Hello @mat-sop! First, I want to thank you for this library; it has helped me a lot in my work, and I would be glad to provide any help in developing.

I found and fixed by myself the same issue as in [PR 256](https://github.com/mirumee/ariadne-codegen/pull/256). Trying to use get_client_settings in a plugin raises a TypeError. Although @slessans has already addressed this issue, I believe adding a test for this fix would be beneficial. I've updated the ExtractOperationsPlugin client test by including a scalar to verify the fix.

If there's anything I can do better, please let me know.

### DIFF
--- a/tests/main/clients/operations/expected_client/__init__.py
+++ b/tests/main/clients/operations/expected_client/__init__.py
@@ -10,6 +10,7 @@ from .custom_operations import (
     C_SUBSCRIPTION_GQL,
     GET_A_GQL,
     GET_A_WITH_FRAGMENT_GQL,
+    GET_S_GQL,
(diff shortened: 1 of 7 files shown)


rafalp:
Hello, thanks for contributing this! I've already added a regression test for #256 in #266, but there's no such thing as too many tests.

Can you please rebase this on main?

alzex3:
Hey @rafalp! Rebased.

rafalp:
Tests exploded, but its same explosion as on main, so I'll ignore this.

PR #279, opened 2024-02-23
Fix str to snake to capture uppercase words followed by an underscore

spicyPug02:
In ariadne_codegen/utils.py, str_to_snake_case was not capturing uppercase words followed by an underscore. \W in the regex would capture all non word characters, but not an underscore. So, things like 'Router_Get' and 'RouterBGP_Get' would both be converted to 'router_get' instead of 'router_bgp_get' in the second case. This caused issues in module naming in the generated clients. This commit fixes the issue by explicitly adding an underscore to the regex.


rafalp (changes requested):
Please rebase on main and add a short description of the change in `CHANGELOG.MD`

spicyPug02:
<code>CHANGELOG.md</code> is updated with a description of the fix

rafalp (approved):

rafalp:
Thank you for contributing!

PR #280, opened 2024-02-28
Refactor `ClientGenerator` class to use private fields

### LINKED ISSUE #277: Field arg named `query` is shadowed by `query` string in generated client methods
When using a query as parameter in a graphQL query, it gets overriden by the code generation query field. As a consequence, all graphQL queries which allow for a query parameter do not work. 

I propose to rename your variable to __query (double underscore ) as of the GraphQL Spec (https://spec.graphql.org/October2021/): 

> Any [Name](https://spec.graphql.org/October2021/#Name) within a GraphQL type system must not start with two underscores "__" unless it is part of the [introspection system](https://spec.graphql.org/October2021/#sec-Introspection) as defined by this specification.

I think this would be a safe choice.

See an illustrative example below, where the argument gets directly overridden and as a consequence, the payload has syntax errors

```python
    async def devices(
        self,
        query: Union[Optional[str], UnsetType] = UNSET,
        page: Union[Optional[int], UnsetType] = UNSET,
        page_size: Union[Optional[int], UnsetType] = UNSET,
        id: Union[Optional[str], UnsetType] = UNSET,
        ordering: Union[Optional[str], UnsetType] = UNSET,
        scope: Union[Optional[Any], UnsetType] = UNSET,
        **kwargs: Any
    ) -> Devices:
        query = gql(
            """
            ...
            ...
            ...

```

Minister944:
ClientGenerator:
- Changed `query` to `_query`.
- Changed `variables` to `_variables`.
- Changed `response` to `_response`.
- Changed `data` to `_data`.
Fixes #277


rafalp (changes requested):
Looking at the diff this change introduced, I am wondering how much extra work would it be to instead detect when one of args is `query` and only then use `_query`. Likewise do same for `_variables`, `_result` and `_data`. 🤔

Minister944:
> Looking at the diff this change introduced, I am wondering how much extra work would it be to instead detect when one of args is `query` and only then use `_query`. Likewise do same for `_variables`, `_result` and `_data`. 🤔

we would have to pass arguments  or set flags to code-generating functions just to determine whether to add `_`
this does not look complicated, but makes the code less readable and increases the level of difficulty of debugging. what would be the purpose of this?

rafalp:
@Minister944 

> what would be the purpose of this?

To don't hit people with pull requests that change 9k places when they update their codegen version.

Minister944:
we solved it in a different way https://github.com/mirumee/ariadne-codegen/pull/281
I am closing this pr

PR #281, opened 2024-03-04
Escape field arg names if they are conflicting with generated client method variables

### LINKED ISSUE #277: Field arg named `query` is shadowed by `query` string in generated client methods

rafalp:
Fixes #277

### DIFF
--- a/ariadne_codegen/client_generators/client.py
+++ b/ariadne_codegen/client_generators/client.py
@@ -122,24 +122,6 @@ class ClientGenerator:
             module = self.plugin_manager.generate_client_module(module)
         return module
 
-    def transform_variable_name(self, variable: str, arguments: ast.arguments):
-        for _args in arguments.args:
-            if _args.arg == variable:
-                return "_" + variable
-        return variable
-
-    def prepare_variable_names(self, arguments: ast.arguments) -> Dict[str, str]:
-        mapped_variable_names = [
-            self._operation_str_variable,
-            self._variables_dict_variable,
-            self._response_variable,
-            self._data_variable,
-        ]
-        variable_names = {}
-        for variable in mapped_variable_names:
-            variable_names[variable] = self.transform_variable_name(variable, arguments)
-        return variable_names
-
     def add_method(
         self,
         definition: OperationDefinitionNode,
@@ -154,7 +136,7 @@ class ClientGenerator:
             definition.variable_definitions
         )
 
-        variable_names = self.prepare_variable_names(arguments)
+        variable_names = self.get_variable_names(arguments)
 
         operation_name = definition.name.value if definition.name else ""
         if definition.operation == OperationType.SUBSCRIPTION:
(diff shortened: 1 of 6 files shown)


ariadne_codegen/client_generators/client.py
```diff
@@ -181,6 +187,23 @@ def add_method(
...
+            self._variables_dict_variable,
+            self._response_variable,
+            self._data_variable,
+        ]
+        variable_names = {}
+        argument_names = set(arg.arg for arg in arguments.args)
+
+        for variable in mapped_variable_names:
+            variable_names[variable] = (
+                "_" + variable if variable in argument_names else variable
```
rafalp, line 202:
```suggestion
                f"_{variable}" if variable in argument_names else variable
```

ariadne_codegen/client_generators/client.py
```diff
@@ -205,9 +229,11 @@ def _generate_subscription_method_def(
                 value=generate_name(ASYNC_ITERATOR), slice_=generate_name(return_type)
             ),
             body=[
-                self._generate_operation_str_assign(operation_str, 1),
-                self._generate_variables_assign(arguments_dict, 2),
-                self._generate_async_generator_loop(operation_name, return_type, 3),
+                self._generate_operation_str_assign(variable_names, operation_str, 1),
```
rafalp, line 232: We now have inconsistency where some functions get `variable_names` as first arg, and some as last arg. I guess we should move it to last arg everywhere because that makes more sense in other places.

Minister944: only 3 functions have `variable_name` as last argument:
https://github.com/mirumee/ariadne-codegen/blob/6438dfb7dc0fe117ee69a444874b6dc66569a627/ariadne_codegen/client_generators/client.py#L148
https://github.com/mirumee/ariadne-codegen/blob/6438dfb7dc0fe117ee69a444874b6dc66569a627/ariadne_codegen/client_generators/client.py#L159
https://github.com/mirumee/ariadne-codegen/blob/6438dfb7dc0fe117ee69a444874b6dc66569a627/ariadne_codegen/client_generators/client.py#L169
Would it not be better to change there to first argument? 
Some [functions](https://github.com/mirumee/ariadne-codegen/blob/6438dfb7dc0fe117ee69a444874b6dc66569a627/ariadne_codegen/client_generators/client.py#L300-L302) have default values in the arguments,  which makes it difficult to add arguments as last.

rafalp: Lets leave it as it is then.

PR #287, opened 2024-03-13
Add plugin to reduce client imports

bombsimon:
* Convert input types and return types to `ast.Constant`
* Import `TYPE_CHECKING` and import types if set
* Import required types inside each method
* Add tests
* Update CHANGELOG
* Update README

---

There are _two_ generated clients to test this with as many permutations as possible both with and without `ShorterResultsPlugin` because it had a major effect on the result (and I had some bugs along the way). However the actual code is contained solely in `ariadne_codegen/contrib`.

This addresses one of the biggest issues from #233. Previously it was very slow to load the `clients.py` because _all_ types used as input and return values were imported. This adds a plugin that converts them to `ast.Constant` and only imports the types if we're in a `TYPE_CHECKING` context.

Example from a project with a couple of hundred methods:

**Before**
```python
>>> import time
>>> def load():
...     now = time.time()
...     import graphql_client.client
...     print(f"took {time.time() - now}")
...
>>> load()
took 5.043132305145264
```

**After**
```python
>>> import time
>>> def load():
...     now = time.time()
...     import graphql_client.client
...     print(f"took {time.time() - now}")
...
>>> load()
took 0.14772796630859375
```

I wouldn't say it solves the issue because even if this improves imports of the client Pydantic v2 is still very slow. Loading some files with many classes can still take several seconds. I don't think we can solve this without rolling back to Pydantic v1 or wait for an upstream fix though.


ariadne_codegen/contrib/no_global_imports.py
```diff
@@ -0,0 +1,359 @@
...
+        )
+
+        module.body.insert(len(non_empty_imports), import_if_type_checking)
+
+        # Import `TYPE_CHECKING`.
+        module.body.insert(
+            len(non_empty_imports),
+            ast.ImportFrom(
+                module="typing",
+                names=[ast.Name("TYPE_CHECKING")],
```
bombsimon, line 328: Should probably use constants for these.

ariadne_codegen/contrib/no_global_imports.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # We add the class to our set of imported in methods - these classes
+        # don't need to be imported at all in the global scope.
+        self.imported_in_method.add(import_class.id)
+        method_def.body.insert(
+            0,
+            ast.ImportFrom(
+                module=self.imported_classes[import_class_id],
+                names=[import_class],
+            ),
```
bombsimon, line 172: I developed this plugin outside of `ariadne-codegen` so I didn't use any helpers to generate these. Tbh I don't think it helps, it just makes it tricker to read.

Let me know if you'd rather me to import and use all the generators even for plugins or if I can keep this as is.

rafalp (commented):
Just few nits for now.

ariadne_codegen/contrib/no_global_imports.py
```diff
@@ -0,0 +1,365 @@
...
+
+from graphql import GraphQLSchema
+
+from ariadne_codegen import Plugin
+
+TYPE_CHECKING_MODULE: str = "typing"
+TYPE_CHECKING_FLAG: str = "TYPE_CHECKING"
+
+
+class NoGlobalImportsPlugin(Plugin):
```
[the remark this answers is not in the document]
bombsimon: Renamed!

README.md
```diff
@@ -96,6 +96,8 @@ Ariadne Codegen ships with optional plugins importable from the `ariadne_codegen
 
 - [`ariadne_codegen.contrib.extract_operations.ExtractOperationsPlugin`](ariadne_codegen/contrib/extract_operations.py) - This extracts query strings from generated client's methods into separate `operations.py` module. It also modifies the generated client to import these definitions. Generated module name can be customized by adding `operations_module_name="custom_name"` to the `[tool.ariadne-codegen.operations]` section in config. Eg.:
 
+- [`ariadne_codegen.contrib.client_forward_refs.ClientForwardRefsPlugin`](ariadne_codegen/contrib/client_forward_refs.py) - This plugin changes generated client module moving all Pydantic models imports under the `TYPE_CHECKING` condition, making them forward references. This greatly improves the import performance for `client` module.
```
rafalp, line 99:
```suggestion
- [`ariadne_codegen.contrib.client_forward_refs.ClientForwardRefsPlugin`](ariadne_codegen/contrib/client_forward_refs.py) - This plugin changes generated client module moving all Pydantic models imports under the `TYPE_CHECKING` condition, making them forward references. This greatly improves the import performance of the `client` module.
```

ariadne_codegen/contrib/__init__.py
```diff
@@ -1,5 +1,11 @@
+from .client_forward_refs import ClientForwardRefsPlugin
 from .extract_operations import ExtractOperationsPlugin
 from .no_reimports import NoReimportsPlugin
 from .shorter_results import ShorterResultsPlugin
 
-__all__ = ["ExtractOperationsPlugin", "NoReimportsPlugin", "ShorterResultsPlugin"]
+__all__ = [
```
rafalp, line 6: Please sort this list alphabetically.

ariadne_codegen/contrib/client_forward_refs.py
```diff
@@ -0,0 +1,365 @@
+"""
+Plugin to only import types for GraphQL responses when you call methods.
```
rafalp, line 2:
```suggestion
Plugin that delays imports of Pydantic models in client module.

Puts all imports under the `typing.TYPE_CHECKING` flag, making
type annotations for generated client's methods forward references.

This greatly improves import time of generated `client` module when
there are many Pydantic models.

Because generated client's methods need type definitions for models
they are using, those models imports will be also inserted in their
bodies.
```

rafalp (changes requested):

rafalp:
More nits but I think this will be good to merge after those.

ariadne_codegen/contrib/__init__.py
bombsimon: Sorted!

ariadne_codegen/contrib/client_forward_refs.py
```diff
@@ -0,0 +1,375 @@
...
+        # TYPE_CHECKING` import block.
+        return_types_not_used_as_input = set(self.input_and_return_types)
+
+        # The ones we import in the method don't need to be imported at all -
+        # unless that's the type we return. This behaviour can differ if you use
+        # a plugin such as `ShorterResultsPlugin` that will import a type that
+        # is different from the type returned.
+        return_types_not_used_as_input.update(
+            {k for k in self.imported_in_method if k not in self.input_and_return_types}
+        )
```
DamianCzajkowski, line 284:
```suggestion
        return_types_not_used_as_input |= self.imported_in_method - self.input_and_return_types
```

ariadne_codegen/contrib/client_forward_refs.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        if not isinstance(call.func, ast.Attribute):
+            return None
+
+        if not isinstance(call.func.value, ast.Name):
+            return None
+
+        return call.func.value
+
+    def _update_imports(self, module: ast.Module):
```
DamianCzajkowski, line 254: Did You consider splitting this function into a smaller ones? Instead of big comments in the middle of the function we could have a smaller functions with docstring that can reduce the complexity of this code. Those two `for` loops looks like they can be separated. Just a suggestion :D

bombsimon: Are you referring to the comment about why we need to be careful wit the `ImportFrom` nodes? I didn't consider splitting it further since they both work on the imports but can absolutely do that. One loop removes the forward refs and the other one adds them.  It would be easy to split by that logic but I would keep the comment anyway since it adds valuable context to why we do it.

bombsimon: Split it into two methods, moved some comments to docstrings

DamianCzajkowski (approved):
LGTM!

rafalp (approved):

PR #292, opened 2024-04-16
Fix: Include directive result type when using `convert_to_snake_case` option

### LINKED ISSUE #272: Incorrect default value for queries with `@include` directive
First of all, I loved the project. It has everything that anyone can expect.

Recently I ran into an issue, When the field name is in camelCase and uses `@include` directive the generated pydantic code doesn't generate proper default value (`None`)

I have used https://graphqlzero.almansi.me/api schema for testing purpose

`queries.graphql`:
```gql
query foo{
  postData: post (id: 1) @include(if: true)  {
    id
  }
}
```

generated pydantic code:
```py
from typing import Optional

from pydantic import Field

from .base_model import BaseModel


class Foo(BaseModel):
    post_data: Optional["FooPostData"] = Field(alias="postData")


class FooPostData(BaseModel):
    id: Optional[str]
```

`post_data` should be `Field(None, alias="postData")` because it's optional.

salmannotkhan:
Fixes #272


ori-lb:
Thanks @salmannotkhan will take a look at this ASAP

ariadne_codegen/client_generators/result_types.py
```diff
@@ -405,6 +406,11 @@ def _process_field_implementation(
         if is_union(field_implementation.annotation):
             keywords[DISCRIMINATOR_KEYWORD] = generate_constant(TYPENAME_ALIAS)
 
+        if keywords and isinstance(field_implementation.value, ast.Constant):
```
DamianCzajkowski, line 409: One thing to ask here, why are you adding this default only if the keywords is not empty? There should be some dependency on default when adding it to the keyword? If yes, can you leave a comment in the code?

salmannotkhan: Because we add the `Field` call only when keyword is not empty.  If there is no `Field` then there is no need for `default` keyword arg that goes into `Field`.

ref: https://github.com/mirumee/ariadne-codegen/pull/292/files#diff-9042b649db08641441c03575181ff29e2d7b821c5de5cf07254fc4845bca9509R414

salmannotkhan: `Field` is only added when property has alias.

DamianCzajkowski (approved):

rafalp:
LGTM but please update the PR to include a `CHANGELOG` entry.

rafalp (approved):

PR #293, opened 2024-05-13
Added deprecation warning output if a deprecation reason exists

AdunSG:
Writing deprecation warning to stdout, if a field used in a query is having a deprecation_reason.

### DIFF
--- a/ariadne_codegen/client_generators/result_types.py
+++ b/ariadne_codegen/client_generators/result_types.py
@@ -1,4 +1,5 @@
 import ast
+import sys
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


AdunSG:
**UseCase:**
I pull the schema using introspection query as already provided by ariadne-codegen.
Now i'd like my pipeline doing that in production to fail, if i'm generating code for a query, which has deprecated fields.
This enables me to react fast to a change of the graphQL API of the host service i'm using.

**Remark:**
For this i'm willing to parse the stdout of the ariadne-codegen for generated DeprecationMessages.
But of course the neatest way would be having a ariade-codegen setting ``fail_on_deprecation``.
If you like and support this idea, i will create another PR including the new setting.

PR #294, opened 2024-05-28
Fix typo in README

sir-sigurd:


DamianCzajkowski:
That's not a fix here, the backslash is for using the '`' sign directly. If I remove the backslash, markdown will mark the text as a code.

bombsimon:
But `` ` `` ([backtick](https://en.wikipedia.org/wiki/Backtick)) isn't the right character to use here, it should be `'` which is an [apostrophe](https://en.wikipedia.org/wiki/Apostrophe). The correct way is to write`can't`, not ``can`t``.

PR #304, opened 2024-07-26
support async_client with custom operations

DamianCzajkowski:


DamianCzajkowski:
This PR needs to be delayed until the MyPy issues that emerged with the release of Python 3.12.4 are resolved.

PR #312, opened 2024-09-04
Allow setting anything with env vars

spookyuser:


ariadne_codegen/settings.py
```diff
@@ -49,7 +53,21 @@ def __post_init__(self):
         if self.schema_path:
             assert_path_exists(self.schema_path)
 
-        self.remote_schema_headers = resolve_headers(self.remote_schema_headers)
```
DamianCzajkowski, line 52: It looks like a crucial line of code was removed, which is necessary for the logic to work correctly. Please restore it, as its absence is causing functionality issues.

ariadne_codegen/settings.py
```diff
@@ -74,9 +92,10 @@ class ClientSettings(BaseSettings):
     scalars: Dict[str, ScalarData] = field(default_factory=dict)
 
     def __post_init__(self):
+        super().__post_init__()
```
DamianCzajkowski, line 95: This change is breaking the tests. Please review and fix the issue to ensure the tests pass.

DamianCzajkowski (changes requested):
This code has caused several tests to fail, and it seems that the validation logic was altered unintentionally. Please revert the changes and rerun the tests to ensure they pass. Once the tests are stable, you can re-submit the pull request for review.

DamianCzajkowski:
Could you also please add a description of what this pull request introduces? It would be helpful to include examples and clarify the specific issues it resolves.

PR #327, opened 2024-11-15
fix: Argument type only uses scalar and ignore list

### LINKED ISSUE #321: GraphQLClientGraphQLMultiError: List dimension mismatch on variable $input_0 and argument input (StagedUploadInput! / [StagedUploadInput!]!)
It does not handle `List` input type correctly

with 

```toml
plugins = ["ariadne_codegen.contrib.extract_operations.ExtractOperationsPlugin"]
```

in my `schema.graphql` I have mutation like this

```gql
type Mutation {
...
  stagedUploadsCreate(input: [StagedUploadInput!]!): StagedUploadsCreatePayload
...
```

which generates `Mutation.staged_uploads_create`  class like this:

```python
    @classmethod
    def staged_uploads_create(
        cls, input: StagedUploadInput
    ) -> StagedUploadsCreatePayloadFields:
        arguments: Dict[str, Dict[str, Any]] = {
            "input": {"type": "StagedUploadInput!", "value": input} # This input.type is incorrect 
            # "input": {"type": "[StagedUploadInput!]!", "value": input} This type is correct
        }
        cleared_arguments = {
            key: value for key, value in arguments.items() if value["value"] is not None
        }
        return StagedUploadsCreatePayloadFields(
            field_name="stagedUploadsCreate", arguments=cleared_arguments
        )
```

then, when executing it sends wrong type. Correct should be `[StagedUploadInput!]!`

```gql
mutation stagedUploadsCreate($input_0: StagedUploadInput!) {
  stagedUploadsCreate(input: $input_0) {
    stagedTargets {
      url
      resourceUrl
      parameters {
        name
        value
      }
    }
  }
}
```

flonou:
This should fix #321 
I'm not completely sure all GraphQL Types are correctly handled

I will check if I can provide tests for this if needed

### DIFF
--- a/ariadne_codegen/client_generators/custom_arguments.py
+++ b/ariadne_codegen/client_generators/custom_arguments.py
@@ -5,6 +5,7 @@ from graphql import (
     GraphQLEnumType,
     GraphQLInputObjectType,
     GraphQLInterfaceType,
+    GraphQLList,
     GraphQLNonNull,
     GraphQLObjectType,
     GraphQLScalarType,
@@ -93,7 +94,7 @@ class ArgumentGenerator:
                 return_arguments_values,
                 arg_name,
                 name,
-                final_type,
+                arg_value.type,
                 used_custom_scalar,
             )
 
@@ -124,7 +125,7 @@ class ArgumentGenerator:
         return_arguments_values: List[ast.expr],
         arg_name: str,
         name: str,
-        final_type: Union[GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType]
+        complete_type: Union[GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType],
         used_custom_scalar: Optional[str],
     ) -> None:
         """Accumulates return arguments."""


ariadne_codegen/client_generators/custom_arguments.py
```diff
@@ -125,12 +124,11 @@ def _accumulate_return_arguments(
...
         arg_name: str,
         name: str,
-        final_type: Union[GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType],
-        is_required: bool,
+        final_type: Union[GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType]
         used_custom_scalar: Optional[str],
     ) -> None:
         """Accumulates return arguments."""
-        constant_value = f"{final_type.name}!" if is_required else final_type.name
+        constant_value = self._generate_complete_type_name(complete_type)
```
normanre, line 131: this would cause a syntax error. the variable `complete_type` does not seem to be defined. I think you can drop the parameter `final_type` and put your new parameter `complete_type` there, then in `generate_arguments()` you pass `complete_type=arg_value.type` to `_accumulate_return_arguments()`.

normanre (changes requested):
Hi,
i think you put up a neat solution. It seems like you did not get to run you code, because there would be a syntax error.

I commented a solution in the line with the error.

Furthermore, a similar problem which is fixed #347 is also missing here.

ariadne_codegen/client_generators/custom_arguments.py
flonou: Indeed I must have messed up with my git client. My local code is exactly what you describe. I'll push a new commit

PR #331, opened 2024-12-25
chore: Pin dependencies to make CI pass

bombsimon:
This fixes spelling of "behavior". I actually don't know where this is coming from or why the test failed but they failed both locally [and in CI](https://github.com/mirumee/ariadne-codegen/actions/runs/12496640540/job/34868553516?pr=330) so this will change the spelling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
 dev = [
   "ariadne",
   "pytest",
-  "pylint",
+  "pylint==3.2.7",
   "mypy",
   "types-toml",
   "pytest-mock",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
   "click~=8.1",
   "graphql-core>=3.2.0,<3.3",
   "toml~=0.10",
-  "httpx~=0.23",
+  "httpx<=0.27.0",
   "pydantic>=2.0.0,<3.0.0",
   "black",
   "isort",
@@ -41,7 +41,7 @@ dev = [
   "types-toml",
   "pytest-mock",
   "pytest-asyncio",
-  "pytest-httpx",
+  "pytest-httpx<=0.34.0",
   "freezegun",
   "requests-toolbelt",
 ]
--- a/ariadne_codegen/contrib/client_forward_refs.py
+++ b/ariadne_codegen/contrib/client_forward_refs.py
@@ -271,7 +271,7 @@ class ClientForwardRefsPlugin(Plugin):
         return_types_not_used_as_input = set(self.input_and_return_types)
 
         # The ones we import in the method don't need to be imported at all -
-        # unless that's the type we return. This behaviour can differ if you use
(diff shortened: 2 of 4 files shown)


bombsimon:
Apparently not as easy, due to the lax pinning `httpx` is now using a version that has [deprecated the app argument](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0280-28th-november-2024). Moving to the new `transport` isn't just a name change since the app used to be the `ariadne.GraphQL` app. 

Pinning is also not a oneliner since the dependency is used by `pytest-httpx`.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
   "click~=8.1",
   "graphql-core>=3.2.0,<3.3",
   "toml~=0.10",
-  "httpx~=0.23",
+  "httpx<=0.27.0",
```
bombsimon, line 28: Pinned due to `app` being [deprecated in 0.28.0](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0280-28th-november-2024)

pyproject.toml
```diff
@@ -36,12 +36,12 @@ dependencies = [
...
   "ariadne",
   "pytest",
-  "pylint",
+  "pylint==3.2.7",
   "mypy",
   "types-toml",
   "pytest-mock",
   "pytest-asyncio",
-  "pytest-httpx",
+  "pytest-httpx<=0.34.0",
```
bombsimon, line 44: Pinned due to [requiring `httpx==0.28.0` since 0.35.0](https://github.com/Colin-b/pytest_httpx/blob/master/CHANGELOG.md#0350---2024-11-28)

pyproject.toml
```diff
@@ -36,12 +36,12 @@ dependencies = [
 dev = [
   "ariadne",
   "pytest",
-  "pylint",
+  "pylint==3.2.7",
```
bombsimon, line 39: Pinned due to [a bunch of fixes and changes in 3.3.0](https://github.com/pylint-dev/pylint/releases/tag/v3.3.0)

bombsimon:
I had to pin 3 dependencies to get this PR green but I wonder if we should take a bigger stab and pin everything. Feels like it's been breaking a few times and CI hasn't been green in 3 months.

For now, I only pinned the bare minimum but feel free to use this branch to pin everything.

I wonder if while we're at it we should also create som setup instructions. I used a bare `virtualenv` locally for a while but am used to `poetry`. Have been trying out `rye` as well but I don't have any preference other than something to both manage the dependencies but also a oneliner to setup everything and install dependencies.

I also suggest moving to [`ruff`](https://docs.astral.sh/ruff) to replace `pylint` and `isort`. They're also in the progress of implementing a type checker to replace `mypy`. 

Let me know if you want any assistance with this or prefer to not do anything or do your own setup!

mjpieters:
Instead of pinning the dependency, just fix the underlying issue with how httpx is used. See #336

mjpieters:
The spelling change comes from the graphql-core package. See #334

PR #347, opened 2025-02-03
better handle list types

matthew-emw:
where our GraphQL schema has a list field as follows:
```graphql
accounts(
    # ...
    """sort the rows by one or more columns"""
    order_by: [accounts_order_by!]
    # ...
  )
```
we were observing generated code with non-list arguments like:
```python
def accounts(
        cls,
        *,
        # ...
        order_by: Optional[accounts_order_by] = None,
        # ...
    )
```

this is pehaps related to issue #321 and PR #327.

my PR reworks this to result in:

```python
def accounts(
        cls,
        *,
        # ...
        order_by: Optional[List[accounts_order_by]] = None,
        # ...
    )
```

i'd like to extend my PR with tests, but:
- i see ArgumentGenerator is currently not covered by any tests
- in `custom_operations.py` i see both an `ArgumentsGenerator` being injected and an `ArgumentGenerator` (no ess) being instantiated
- does this imply a _direction of travel_ and that `ArgumentsGenerator` should be used going forward?

i'd be grateful for any guidance. thanks!

### DIFF
--- a/ariadne_codegen/client_generators/custom_arguments.py
+++ b/ariadne_codegen/client_generators/custom_arguments.py
@@ -6,6 +6,7 @@ from graphql import (
     GraphQLInputObjectType,
     GraphQLInterfaceType,
     GraphQLNonNull,
+    GraphQLList,
     GraphQLObjectType,
     GraphQLScalarType,
     GraphQLUnionType,
@@ -23,6 +24,7 @@ from ..codegen import (
     generate_dict,
     generate_import_from,
     generate_keyword,
+    generate_list_annotation,
     generate_name,
     generate_subscript,
     generate_tuple,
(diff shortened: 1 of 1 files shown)


ariadne_codegen/client_generators/custom_arguments.py
```diff
@@ -127,10 +131,12 @@ def _accumulate_return_arguments(
         name: str,
         final_type: Union[GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType],
         is_required: bool,
+        is_list: bool,
         used_custom_scalar: Optional[str],
     ) -> None:
         """Accumulates return arguments."""
-        constant_value = f"{final_type.name}!" if is_required else final_type.name
+        constant_value = f"[{final_type.name}]" if is_list else final_type.name
+        constant_value = f"{constant_value}!" if is_required else constant_value
```
normanre, line 139: this will not work for this type of fields `[String!]!` (see #321)

normanre (changes requested):
Hi,

i have the same issue and have fixed it myself.

I think your solution wont work only because of #321.
When you add the solution from #327 your solution would be perfect.

Furthermore, the solution in #327 does have an syntax error. I commented a solution in #327.

DamianCzajkowski:
Fixed in https://github.com/mirumee/ariadne-codegen/pull/363

Issue #231, opened 2023-10-19
Header environment placeholder not compatible with oauth2


gnthibault:
To whom it may concern,
I think there is an issue with the get_header_value method:
https://github.com/mirumee/ariadne-codegen/blob/8bc43888a3e63fcd21cdbca07e2e82c6ca46dbc2/ariadne_codegen/settings.py#L246

In particular, the value.startswith won't work in case you have the following type of config for oauth2:
remote_schema_headers = {"Authorization" = "Bearer $AUTH_TOKEN"} # extra headers that are passed along with introspection query

bombsimon:
As a workaround you can do something like
`export BEARER_AUTH_TOKEN="Bearer $AUTH_TOKEN"` and use the variable `$BEARER_AUTH_TOKEN` instead.

Another alternative is to extend the `BaseClient` and set the header yourself, this is what I've done.

gnthibault:
proposed naive version:

def get_header_value(value: str) -> str:
    env_var_prefix = "$"
    l_keys = value.strip().split(sep=" ")
    compiled_keys = []
    for key in l_keys:
        token_pos = key.rfind(env_var_prefix)
        if token_pos >= 0 and token_pos<len(key)-1:
            env_var_name = key[token_pos+1:]
            var_value = os.environ.get(env_var_name)
            if not var_value:
                raise InvalidConfiguration(
                    f"Environment variable {env_var_name} not found."
                )
            else:
                key = key[:token_pos] + var_value
        compiled_keys.append(key)

rafalp:
@bombsimon solution is the way to do it. We could add the note in the docs that we are not doing any complex parsing on setting values.

mat-sop:
@bombsimon This utility is used only for headers attached to remote schema introspection query (`remote_schema_headers` config option). Extending `BaseClient` won't have an effect here, so 1st approach with `export ...` is the way to do it.

gnthibault:
Ok good for me, I will close the issue

PR #349, opened 2025-02-11
Add support for embedded variables in headers

zyv:
Hey,

Thanks for this awesome generator!

Unfortunately, I can't use it to fetch GitHub's own schemas because they require `Authorization` headers in the form of `Bearer token`. Currently, the generator only supports variable expansion if it's the only element in the header value. However, I can't hardcode the token in my `pyproject.toml` to work around this limitation.

So I added support for embedded and multiple variables (not that I need multiple variables, but you basically get that for free with my approach, so why not...).  There may be situations where it won't work like shell, but I would prefer simplicity over handling all possible corner cases.

Hope this helps you and others!

### DIFF
--- a/ariadne_codegen/settings.py
+++ b/ariadne_codegen/settings.py
@@ -1,5 +1,6 @@
 import enum
 import os
+import re
 from dataclasses import dataclass, field
 from keyword import iskeyword
 from pathlib import Path
@@ -278,16 +279,17 @@ def resolve_headers(headers: Dict) -> Dict:
 
 
 def get_header_value(value: str) -> str:
-    env_var_prefix = "$"
-    if value.startswith(env_var_prefix):
-        env_var_name = value.lstrip(env_var_prefix)
+    replacements = {}
+    for env_var_name in re.findall(r"\$([A-z][A-z0-9_]*)", value):
         var_value = os.environ.get(env_var_name)
         if not var_value:
             raise InvalidConfiguration(
                 f"Environment variable {env_var_name} not found."
             )
-        return var_value
-
+        replacements[f"${env_var_name}"] = var_value
+    if replacements:
+        pattern = re.compile("|".join(map(re.escape, replacements.keys())))
+        return pattern.sub(lambda match: replacements[match.group(0)], value)
     return value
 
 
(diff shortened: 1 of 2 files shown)


bombsimon:
This looks like a duplicate of #332 

Did you consider storing `Bearer $YOUR_TOKEN` in a variable instead? See https://github.com/mirumee/ariadne-codegen/issues/231#issuecomment-1770433611

zyv:
> This looks like a duplicate of #332

Sorry about that! I should have searched for other PRs before opening mine.

As far as I'm concerned, it doesn't really matter which PR gets merged. I'll take a look at your code.

> Did you consider storing `Bearer $YOUR_TOKEN` in a variable instead? See [#231 (comment)](https://github.com/mirumee/ariadne-codegen/issues/231#issuecomment-1770433611)

Yes, I did - but it's an ugly and clumsy hack. All tools (GitHub CLI, SDKs, etc.) standardize on `GITHUB_TOKEN`, and if I set `GITHUB_TOKEN` to `Bearer ...` it will break them. I can only set a second variable specifically for Ariadne like `export ARIADNE_GITHUB_TOKEN="Bearer $GITHUB_TOKEN"`.

But what's worse is the very annoying and confusing default behavior. I've never seen a tool that claims to support "variables", but only resolves a variable if it's the only content of the string. This is not documented and extremely counterintuitive.

In fact, it's so confusing that I didn't even know what to search for and ended up debugging the code to figure out why I was getting 401 or 403 no matter what I tried. At that point, implementing proper variable support, including tests, only took a quarter of an hour, and I didn't bother looking for issues.

So if you decide to reject the PR, please at least document the weird notion of the variable directly where variables are mentioned, to save others searching and re-implementing the same.

ariadne_codegen/settings.py
```diff
@@ -278,16 +279,17 @@ def resolve_headers(headers: Dict) -> Dict:
 
 
 def get_header_value(value: str) -> str:
-    env_var_prefix = "$"
-    if value.startswith(env_var_prefix):
-        env_var_name = value.lstrip(env_var_prefix)
+    replacements = {}
+    for env_var_name in re.findall(r"\$([A-z][A-z0-9_]*)", value):
         var_value = os.environ.get(env_var_name)
```
zyv, line 284:
```suggestion
    for env_var_name in re.findall(r"\$([A-z][A-z0-9_]*|{[A-z][A-z0-9_]*})", value):
        var_value = os.environ.get(env_var_name.removeprefix("{").removesuffix("}"))
```
... to support `${VARIABLES}` if desired.

Issue #355, opened 2025-02-18
websockets.client.connect is deprecated in async_base_client.py


Liavshab:
We are encountering a DeprecationWarning due to the use of websockets.client.connect in async_base_client.py, which is deprecated in recent versions of the websockets library.

```
<project_dir>/generated/async_base_client.py:21
 <project_dir>/generated/async_base_client.py:21: DeprecationWarning: websockets.client.connect is deprecated
    from websockets.client import (  # type: ignore[import-not-found,unused-ignore]
``` 

Steps to Reproduce:
	1.	Run the code with the latest version of websockets installed.
	2.	Observe the deprecation warning in logs.

Proposed Fix:
Modify the import statement in async_base_client.py from:

```python
from websockets.client import (  # type: ignore[import-not-found,unused-ignore]
    WebSocketClientProtocol,
    connect as ws_connect,
)
``` 
to
```python
from websockets.sync.client import (  # type: ignore[import-not-found,unused-ignore]
    WebSocketClientProtocol,
    connect as ws_connect,
)
```

PR #336, opened 2025-01-08
tests: Use an explicit transport for httpx

mjpieters:
httpx deprecated the `app` shortcut in 0.27 and removed it altogether in 0.28. Use the recommended `ASGITransport` transport to replace it.

Fixes #335

PR #357, opened 2025-02-19
fix: update websockets

mgradalska:
Fixes https://github.com/mirumee/ariadne-codegen/issues/355

After upgrading `websockets`, application was throwing deprecation warnings. These changes fix the issue, but require `websockets` to be upgraded to at least 14.2 version where `ClientConnection` was introduced. 

~~Also, fixed deprecated `app` argument in httpx `AsyncClient`.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,5 @@ dmypy.json
 
 # System files
 .DS_Store
+
+.idea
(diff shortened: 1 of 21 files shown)


bombsimon:
> Also, fixed deprecated `app` argument in httpx `AsyncClient`.

Nice! However the quoted part seems like a duplicate of #336. Not that the maintainers seem to actively look at the PRs but decreasing scope and isolate each PR to a single topic is probably a good idea

mgradalska:
I thought no one was working on `httpx` issue, I missed that PR. Anyways, you're right. I removed `httpx` related changes.

PR #358, opened 2025-02-22
Fix typo in `settings.py` message text

zyv:
Hopefully, *cop**y**ing*, not just *coping* with the base class :)


zyv:
Sorry, but mypy failures are not mine :(

Issue #351, opened 2025-02-17
chore: introduce python 3.13, ruff...


drozdowsky:
Issue similar to https://github.com/mirumee/ariadne/pull/1221 - let's unite config, styles, hatch across codegen too.

provide explicit configs for mypy, ruff, coverage
 - Python 3.13 support introduced

What needs to be done:

- make a commit that will fix style issues
- ignore the style commit in .git-blame-ignore-revs
- fix mypy errors
- update GitHub actions
- update CONTRIBUTING.md

PR #359, opened 2025-02-25
Introduce py3.13, ruff, leverage hatch

drozdowsky:
Addresses: https://github.com/mirumee/ariadne-codegen/issues/351

Github actions and pyproject.toml are now similar to what you can see in main Ariadne repo.

Few issues I see:
- currently auto-generated code is formatted with black+isort (used as imported modules) - this could clash with ruff. I suggest moving to ruff there too.
- I imagine someone has to configure pypi with new publish flow


KarolJagodzinski:
Thanks @drozdowsky for the contribution!

Issue #317, opened 2024-09-24
Confusing snake case behaviour


nevoodoo:
Hi!

Just want to say amazing work with the project, and thank you for the Custom Operations, it has been a life saver indeed.

I wanted to get some clarity on an issue we recently noticed:

```python
@strawberry.field
    async def ourdna_dashboard(
        self, info: Info, root: 'Project'
    ) -> 'GraphQLOurDNADashboard':
        connection = info.context['connection']
        ourdna_layer = OurDnaDashboardLayer(connection)
        if not root.id:
            raise ValueError('Project must have an id')
        ourdna_dashboard = await ourdna_layer.query(project_id=root.id)
        # pylint: disable=no-member
        return GraphQLOurDNADashboard.from_internal(ourdna_dashboard)
```

If this is the field definition, the resulting codegen client has the field as `ourdna_dashboard` whereas Strawberry expects it as `ourdnaDashboard` and so the error on the server is:

```
raise GraphQLClientGraphQLMultiError.from_errors_dicts(
graphql_client.exceptions.GraphQLClientGraphQLMultiError: Cannot query field 'ourdna_dashboard' on type 'GraphQLProject'. Did you mean 'ourdnaDashboard'?
```

Happy to help and push a fix PR if I can be nudged in the right direction :)

DamianCzajkowski:
Fixed by #377, the custom-operation field builder now uses the original schema field name (e.g., ourdnaDashboard) for the query, keeping snake_case only for the Python method name. Closing as resolved.

PR #377, opened 2025-09-04
Fix snake cased gql field names

ewen-lorimer:
Updated the custom field generate to use the original field name (extracted from the graphql schema) instead of the converted (snake-cased) field name.

The issue this PR aims to solve was raised in #317, but also happens to solve an issue I've personally been having with ariadne-codegen.


DamianCzajkowski:
Hi @ewen-lorimer! Thanks for the PR and the fix. Sorry for the delay - merging now!

PR #385, opened 2025-09-29
Add a setting that sets default=None to all optional fields so they can be missing from response and still parse

### LINKED ISSUE #309: [feature request] default `= None` for nullable fields that evaluate to `Optional[T]` in Python model
Ariadne has been very useful for our team which uses GraphQL backed by Python services.

When we add new fields to a type, existing records in our underlying database (DynamoDB) do not have the new attribute - so we run into a ValidationError (missing) when creating a Python object with newly generated Pydantic model. The fix for this would be to add the default `= None` for attributes that are `Optional`, which allows creation of the model class instance to continue even if the new attribute is not in the old record. 

Is there already a way to do this that we're not leveraging properly?

```graphql
# In schema.graphql which is our config `schema_path`
type Location {
  locationId: ID!
  parentId: ID
  name: String
}
```

```graphql
# In queries.graphql which is our config `queries_path`
fragment Location on Location {
  locationId
  parentId
  name
}
```

After running codegen we get:
```python
# generated/fragments.py
class Location(BaseModel):
    locationId: str
    parentId: Optional[str]
    name: Optional[str]
```

We'd like:
```python
# generated/fragments.py
class Location(BaseModel):
    locationId: str
    parentId: Optional[str] = None
    name: Optional[str] = None
```

etareduction:
Closes #309

### DIFF
--- a/ariadne_codegen/client_generators/fragments.py
+++ b/ariadne_codegen/client_generators/fragments.py
@@ -20,6 +20,7 @@ class FragmentsGenerator:
         convert_to_snake_case: bool = True,
         custom_scalars: Optional[Dict[str, ScalarData]] = None,
         plugin_manager: Optional[PluginManager] = None,
+        default_optional_fields_to_none: bool = False,
     ) -> None:
         self.schema = schema
         self.enums_module_name = enums_module_name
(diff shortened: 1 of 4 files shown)


DamianCzajkowski:
Hi @etareduction, Thanks for your pull request! Can you add some unit tests to verify your code?

etareduction:
@DamianCzajkowski Sorry i was too busy to come through with that, but my friend here came to help and added a few tests. Does it look good?

PR #391, opened 2025-11-24
Update variable collection in base_operation.py

keyfayaz:
Fix issue where subfield variable does not get copied into formatted vars


DamianCzajkowski:
Ok, I see, this issue was previously fixed, but we missed the tests part, thanks for that! I can do a release 0.16.1 with this fix :D

PR #393, opened 2025-11-27
Fix __typename field handling to use GraphQLString

MattKobayashi:
This PR is just #388 with a fix for the static analysis failure.


DamianCzajkowski:
@MattKobayashi, thanks for bringing it back! Unfortunately, there are still some static analysis failures that need to be addressed. I will wait for you before next release :)

PR #397, opened 2025-12-04
fix TypeError: Redefinition of reserved type 'String'

tubaman:


DamianCzajkowski:
Fixed in https://github.com/mirumee/ariadne-codegen/pull/393

PR #419, opened 2026-03-12
Add models_only option to skip client code generation

magicmark:
## Summary

- Adds a `models_only` configuration option (`models_only = true` in TOML) that generates only Pydantic models (input types, enums, result types, fragments, base model) without the runtime client code (client class, base client, exceptions)
- Makes `queries_path` and base client configuration optional when `models_only` is enabled
- Removes the httpx dependency for type-only use cases

Closes #418

## Test plan

- [x] All existing tests pass (645 passed, 0 failures)
- [x] New tests verify models_only skips client/exceptions files
- [x] New tests verify `__init__.py` has no client imports in models_only mode
- [x] New tests verify result types are still generated from operations
- [x] New settings tests verify models_only works without queries_path and skips base client validation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ariadne_codegen/client_generators/package.py
+++ b/ariadne_codegen/client_generators/package.py
@@ -92,6 +92,7 @@ class PackageGenerator:
         default_optional_fields_to_none: bool = False,
         include_typename: bool = True,
         ignore_extra_fields: bool = True,
+        models_only: bool = False,
     ) -> None:
         self.package_path = Path(target_path) / package_name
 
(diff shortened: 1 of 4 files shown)


DamianCzajkowski:
I’m on board with the goal here, and the use case makes sense.
I’m not sold on adding models_only as a boolean in settings, though.

This feels more like a generation mode than a config toggle. I’d rather model it like existing strategies (client, graphqlschema) and add models_only at the entrypoint/strategy level, then keep settings focused on strategy-specific params.

So +1 on the feature itself. I’d just prefer shipping it as a strategy/mode instead of models_only = true in settings.

amin-farjadi:
Hi @magicmark, I love this idea. Would you like help implementing @DamianCzajkowski comment?

PR #436, opened 2026-06-21
feat: models only mode

amin-farjadi:
Closes issue #418 and addresses comments on PR #419.


amin-farjadi:
hi @DamianCzajkowski would you be able to review this PR?

DamianCzajkowski:
Hi @amin-farjadi, this feature will definitely be merged. We are currently discussing the features and the future of the codegen, so don’t worry; it will be merged this month. Thank you for contributing!

ariadne_codegen/config.py
```diff
@@ -103,6 +108,57 @@ def get_section(config_dict: dict) -> dict:
     raise MissingConfiguration(f"Config has no [{tool_key}.{codegen_key}] section.")
 
 
+def get_models_only_settings(config_dict: dict) -> ModelsOnlySettings:
```
Kwaidan00, line 111: Please correct me if I'm wrong, but this function is a copy of `get_client_settings` with the following changes:
- return typing
- `settings_fields_names` is set based on `ModelsOnlySettings`
- returned object is constructed as `ModelsOnlySettings` instance

I believe there could be a single function which takes the model as parameter and is used for both the client settings and `ModelsOnlySettings`. Then, there could be two facade functions `get_client_settings` and `get_models_only_settings` that returns `foo(config_dict, ModelsOnlySettings)` or `foo(config_dict, ClientSettings)`.

ariadne_codegen/client_generators/package.py
```diff
@@ -460,23 +471,41 @@ def get_package_generator(
...
+    if models_only:
+        base_client_import = generate_import_from(
+            names=["object"], from_="builtins", level=0
+        )
+        client_name = "Client"
+        base_client = "object"
+        async_client = True
+        base_client_file_path = ""
+        base_client_module_name = ""
+        client_file_name = "client"
```
Kwaidan00, line 483: You create a fake objects here just to satisfy the requirement to provide `client_generator` to `PackageGenerator`.

However, the strategy you implemented does not need to know anything about client at all - it's the mode without it.

This part of the code and multiple checks `if not self.models_only` in the generator methods suggest there could be a solution easier to maintain. There should be separate `PackageGenerator` subclasses: for the client mode and models only mode. Key concepts:
- most of the logic can be shared in the base class
- `generate` method, as the most important function that describes the steps to perform, should be implemented in the subclasses
- `add_operation` could live in the base class, the "client mode" class can call `super().add_operation` and then call the additional `self.client_generator.add_method()`
- `_validate_unique_file_names` can live in the base class, but the subclasses should implement a method to return expected file names
- similarly, `_copy_files` can live in the base class, but the subclasses can implement a method to return files to copy
- finally, the `__init__` can share most of the logic, but the "client mode" class can expect additional arguments like `client_generator` and other client-related settings, unused for "models only" mode.

ariadne_codegen/settings.py
```diff
@@ -332,6 +333,95 @@ def used_settings_message(self) -> str:
...
+    include_comments: CommentsStrategy = field(default=CommentsStrategy.STABLE)
+    convert_to_snake_case: bool = True
+    include_all_inputs: bool = True
+    include_all_enums: bool = True
+    files_to_include: list[str] = field(default_factory=list)
+    scalars: dict[str, ScalarData] = field(default_factory=dict)
+    default_optional_fields_to_none: bool = False
+    include_typename: bool = True
+    ignore_extra_fields: bool = True
+    multipart_uploads: bool = True
```
Kwaidan00, line 353: This specific key is used to decide if the base client should contain multipart uploads implementation or not. I think it's the typo to include it here.

ariadne_codegen/settings.py
```diff
@@ -332,6 +333,95 @@ def used_settings_message(self) -> str:
         )
 
 
+@dataclass
+class ModelsOnlySettings(BaseSettings):
```
Kwaidan00, line 337: This model seems to contain a subset of fields defined in `ClientSettings`. Similarly, the `__post_init__` is a shortened copy of `ClientSettings.__post_init__`.

I think better solution is to create a model e.g. `GeneratorSettings(BaseSettings)`, it will contain the common set of fields and the common logic in `__post_init__`. Then, `ClientSettings` and `ModelsOnlySettings` should inherit `GeneratorSettings` and optionally add custom logic and additional fields.

tests/main/graphql_schemas/example/expected_schema.py
```diff
@@ -343,7 +343,6 @@
                 DirectiveLocation.ARGUMENT_DEFINITION,
                 DirectiveLocation.INPUT_FIELD_DEFINITION,
                 DirectiveLocation.ENUM_VALUE,
-                DirectiveLocation.DIRECTIVE_DEFINITION,
```
Kwaidan00, line 346: Why did you drop it?

tests/main/models_only/example/expected_client/base_model.py
```diff
@@ -0,0 +1,28 @@
...
+class BaseModel(PydanticBaseModel):
+    model_config = ConfigDict(
+        populate_by_name=True,
+        validate_assignment=True,
+        arbitrary_types_allowed=True,
+        protected_namespaces=(),
+    )
+
+
+class Upload:
```
Kwaidan00, line 24: This generated example shows the `Upload` class is not used anywhere (in the client mode, it is used in the client module). I think you can easily change your code to not include the `Upload` class by using `base_model_no_upload.py` instead of `base_model.py`.

tests/main/test_main.py
```diff
@@ -243,17 +244,6 @@ def test_main_shows_version():
...
-            (
-                CLIENTS_PATH / "no_multipart_upload" / "pyproject.toml",
-                (
-                    CLIENTS_PATH / "no_multipart_upload" / "queries.graphql",
-                    CLIENTS_PATH / "no_multipart_upload" / "schema.graphql",
-                ),
-            ),
-            "expected_client",
-            CLIENTS_PATH / "no_multipart_upload" / "expected_client",
-        ),
```
Kwaidan00, line 256: Why did you delete this test case?

Kwaidan00 (commented):
Hi @amin-farjadi, thank you for your contribution! I left some comments about the overall implementation shape. Please review and let me know if you are planning to apply those changes.

amin-farjadi:
> Hi @amin-farjadi, thank you for your contribution! I left some comments about the overall implementation shape. Please review and let me know if you are planning to apply those changes.

Thank you for the review. I will make those changes within the next 2-3 days.

PR #437, opened 2026-06-22
feat: multipart upload opt in AR-6 

Minister944:
Introduces a `multipart_uploads` boolean setting (default: true) that,
when set to false, generates lighter base client variants that omit
multipart/file-upload handling. Adds four new no-upload base client
dependency files (async, sync, with/without OpenTelemetry) and a
corresponding integration test fixture.


ariadne_codegen/client_generators/package.py
```diff
@@ -69,6 +70,7 @@ def __init__(
         async_client: bool = True,
         base_client_name: str = "AsyncBaseClient",
         base_client_file_path: str = DEFAULT_ASYNC_BASE_CLIENT_PATH.as_posix(),
+        base_client_module_name: str = "",
```
Kwaidan00, line 77: 1. This class provides the defaults for parameters to generate the functional default implementation (default async base client). Shouldn't this param be set then?
2. What should be set here for using custom base client class?

ariadne_codegen/client_generators/package.py
```diff
@@ -359,28 +364,48 @@ def _generate_fragments(self):
...
-            if not self.ignore_extra_fields and source_path.name == "base_model.py":
-                code = add_extra_to_base_model(code)
             if self.plugin_manager:
                 code = self.plugin_manager.copy_code(code)
             target_path = self.package_path / source_path.name
             target_path.write_text(code)
             self._generated_files.append(target_path.name)
 
+        base_client_target_stem = (
+            self.base_client_module_name or self.base_client_file_path.stem
```
Kwaidan00: The logic like that should probably live in the initializator.

ariadne_codegen/client_generators/package.py
```diff
@@ -359,28 +364,48 @@ def _generate_fragments(self):
...
+        code = self._add_comments_to_code(
+            self.base_model_file_path.read_text(encoding="utf-8")
+        )
+        if not self.ignore_extra_fields:
+            code = add_extra_to_base_model(code)
+        if self.plugin_manager:
+            code = self.plugin_manager.copy_code(code)
+        base_model_target = self.package_path / "base_model.py"
+        base_model_target.write_text(code)
+        self._generated_files.append(base_model_target.name)
```
Kwaidan00: This code is very similar to itself; it also repeats the algorithm (slightly adjusted) from the loop above. 

WDYT to change `files_to_copy` into a mapping "file_path_to_copy: target_file_name"; the files from `self.files_to_include` will use `source_path.name` (as it's implemented in the loop) and for base client and base model it will be set accordingly?

ariadne_codegen/client_generators/package.py
DamianCzajkowski: This will target to `async_base_client_no_upload` not to `async_base_client`
probably fix for that: 
```
from_=settings.base_client_module_name or Path(settings.base_client_file_path).stem,
```

tests/main/clients/no_multipart_upload/expected_client/client.py
```diff
@@ -0,0 +1,47 @@
+# Generated by ariadne-codegen
+# Source: queries.graphql
+
+from typing import Any
+
+from .async_base_client_no_upload import AsyncBaseClient
```
DamianCzajkowski: Twhs import points to a module that is never generated. The base client file is written as `async_base_client` not `async_base_client_no_upload`

tests/main/clients/no_multipart_upload/expected_client/async_base_client.py
```diff
@@ -0,0 +1,308 @@
...
+from collections.abc import AsyncIterator
+from typing import Any, Optional, TypeVar, cast
+from uuid import uuid4
+
+import httpx
+from pydantic import BaseModel
+from pydantic_core import to_jsonable_python
+
+from .base_model import UNSET
+from .exceptions import (
```
DamianCzajkowski, line 15: For some reason, this file is not copied in the no_upload mode.

ariadne_codegen/client_generators/package.py
DamianCzajkowski, line 254: To fix the issue with no generated file with an exceptions in no upload mode, you have to add new paths here

DamianCzajkowski: 258 and 259 will probably validate the wrong file names. From what I see, it will validate the file names with the no_upload suffix, but not the actual generated file names

DamianCzajkowski (changes requested):

ariadne_codegen/client_generators/package.py
Minister944: `base_client_module_name` is the name the base client is written under in the generated package (and what `client.py `imports). It has to stay because PackageGenerator needs that output name, and it can't be derived from `base_client_file_path.stem`: for the default no-upload variants the source file (`async_base_client_no_upload.py`) differs from the output module (`async_base_client.py`). Using the stem would give the wrong name.

So it's a computed value the generator consumes, not user config. I'd keep resolving it together with `base_client_name` / `base_client_file_path` in `_set_default_base_client_data`, since splitting it out produces the same value anyway and just adds a "default vs custom?" branch.

DamianCzajkowski (approved):

Kwaidan00 (approved):

PR #439, opened 2026-06-28
feat: add schema_paths for loading schema from local paths and installed packages AR-3

Minister944:
Introduce a new `schema_paths` setting that builds the GraphQL schema from
multiple sources. Each entry is resolved either as a dotted Python attribute
path (e.g. `pkg.SCHEMA_DIR` or `pkg.get_schema_files`) - useful for pulling
type definitions from installed packages - or as a local file/directory path.

- schema.py: add `resolve_schema_paths` and `get_graphql_schema_from_paths`
- main.py: wire `schema_paths` into both `client` and `graphqlschema` strategies
- settings.py: add the `schema_paths` field and make the three schema sources
  (`schema_path`, `schema_paths`, `remote_schema_url`) mutually exclusive -
  providing more than one now raises InvalidConfiguration; simplify
  `using_remote_schema` accordingly
- docs/README: document `schema_paths` and the mutual-exclusivity rule

CI/CD does not pass. This is an unrelated error that is being fixed as part of [AR-6](https://github.com/mirumee/ariadne-codegen/pull/437)

## Acknowledgements

While working on this, we noticed #431 by @esfomeado, which also tackled
loading the schema from more than one source. After discussing it as a team,
we went with a different approach - a dedicated `schema_paths` setting that
additionally resolves dotted Python attribute paths from installed packages -
which we felt fit the project better. Thanks @esfomeado for raising the problem
and for your contribution! 🙏

Co-authored-by: esfomeado <1906254+esfomeado@users.noreply.github.com>

[AR-6]: https://mirumee.atlassian.net/browse/AR-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ariadne_codegen/schema.py
+++ b/ariadne_codegen/schema.py
@@ -29,10 +29,11 @@ from typing_extensions import Any
 from .client_generators.constants import MIXIN_FROM_NAME, MIXIN_IMPORT_NAME, MIXIN_NAME
 from .exceptions import (
     IntrospectionError,
+    InvalidConfiguration,
     InvalidGraphqlSyntax,
     InvalidOperationForSchema,
 )
-from .settings import IntrospectionSettings
+from .settings import IntrospectionSettings, assert_path_exists
 
 
 def filter_operations_definitions(
@@ -145,39 +146,54 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
 def resolve_schema_paths(sources: list[str]) -> list[Path]:
     """Resolve a list of schema sources to concrete file paths.
 
-    Each entry is tried as a dotted Python import path first (e.g.
-    ``pkg.SCHEMA_DIR`` or ``pkg.get_schema_files``). If the import fails the
-    entry is treated as a local filesystem path instead.
+    Each entry is first tried as a local filesystem path - a directory (searched
+    recursively for graphql files) or a file. If it points to neither, it is
+    treated as a dotted Python import path (e.g. ``pkg.SCHEMA_DIR`` or
+    ``pkg.get_schema_files``); an import path that cannot be resolved raises
+    ``InvalidConfiguration``.
     """
     result: list[Path] = []
     for source in sources:
-        if (
-            "." in source
-            and "/" not in source
-            and not source.endswith((".graphql", ".graphqls", ".gql"))
-        ):
-            try:
-                module_path, attr = source.rsplit(".", 1)
-                module = importlib.import_module(module_path)
-                obj = getattr(module, attr)
-                if callable(obj):
-                    result.extend(Path(f) for f in obj())
-                elif isinstance(obj, (str, Path)):
-                    dir_path = Path(obj)
-                    if dir_path.is_dir():
-                        result.extend(sorted(walk_graphql_files(dir_path)))
-                    else:
-                        result.append(dir_path)
-                continue
-            except (ImportError, ModuleNotFoundError):
-                pass
-
-        local_path = Path(source)
-        if local_path.is_dir():
-            result.extend(sorted(walk_graphql_files(local_path)))
+        path = Path(source)
+        if path.is_dir():
+            result.extend(sorted(walk_graphql_files(path)))
+        elif path.is_file():
+            result.append(path)
         else:
-            result.append(local_path)
+            resolved = _resolve_import_source(source)
+            for resolved_path in resolved:
+                assert_path_exists(resolved_path.as_posix())
+            result.extend(resolved)
     return result
+
+
+def _resolve_import_source(source: str) -> list[Path]:
+    """Resolve a dotted Python import path to schema file paths.
+
+    The imported object may be a callable returning a list of paths, or a string
+    / ``Path`` pointing to a file or a directory.
+    """
+    try:
+        module_path, attr = source.rsplit(".", 1)
+        module = importlib.import_module(module_path)
+        obj = getattr(module, attr)
+    except (ImportError, AttributeError, ValueError) as exc:
+        raise InvalidConfiguration(
+            f"Could not resolve schema source '{source}'. It is neither an "
+            f"existing file/directory nor an importable attribute: {exc}."
+        ) from exc
+
+    if callable(obj):
+        return [Path(f) for f in obj()]
+    if isinstance(obj, (str, Path)):
+        obj_path = Path(obj)
+        if obj_path.is_dir():
+            return sorted(walk_graphql_files(obj_path))
+        return [obj_path]
+    raise InvalidConfiguration(
+        f"Schema source '{source}' resolved to {type(obj).__name__}; expected a "
+        "path string, a Path, or a callable returning a list of paths."
+    )
 
 
 def get_graphql_schema_from_paths(schema_paths: list[str]) -> GraphQLSchema:
--- a/ariadne_codegen/schema.py
+++ b/ariadne_codegen/schema.py
@@ -194,15 +194,10 @@ def _resolve_import_source(source: str) -> list[Path]:
 
     if callable(obj):
         return [Path(f) for f in obj()]
-    if isinstance(obj, (str, Path)):
-        obj_path = Path(obj)
-        if obj_path.is_dir():
-            return sorted(walk_graphql_files(obj_path))
-        return [obj_path]
-    raise InvalidConfiguration(
-        f"Schema source '{source}' resolved to {type(obj).__name__}; expected a "
-        "path string, a Path, or a callable returning a list of paths."
-    )
+    obj_path = Path(obj)
+    if obj_path.is_dir():
+        return sorted(walk_graphql_files(obj_path))
+    return [obj_path]
 
 
 def get_graphql_schema_from_paths(schema_paths: list[str]) -> GraphQLSchema:
--- a/ariadne_codegen/schema.py
+++ b/ariadne_codegen/schema.py
@@ -1,3 +1,4 @@
+import importlib
 from collections.abc import Generator, Sequence
 from dataclasses import asdict
 from pathlib import Path
@@ -136,6 +137,51 @@ def introspect_remote_schema(
 def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
     """Get graphql schema build from provided path."""
     schema_str = load_graphql_files_from_path(Path(schema_path))
+    graphql_ast = parse(schema_str)
+    schema: GraphQLSchema = build_ast_schema(graphql_ast, assume_valid=True)
+    return schema
+
+
+def resolve_schema_paths(sources: list[str]) -> list[Path]:
+    """Resolve a list of schema sources to concrete file paths.
+
+    Each entry is tried as a dotted Python import path first (e.g.
+    ``pkg.SCHEMA_DIR`` or ``pkg.get_schema_files``). If the import fails the
+    entry is treated as a local filesystem path instead.
+    """
+    result: list[Path] = []
+    for source in sources:
+        if "." in source and "/" not in source and not source.endswith(
+            (".graphql", ".graphqls", ".gql")
+        ):
+            try:
+                module_path, attr = source.rsplit(".", 1)
+                module = importlib.import_module(module_path)
+                obj = getattr(module, attr)
+                if callable(obj):
+                    result.extend(Path(f) for f in obj())
+                elif isinstance(obj, (str, Path)):
+                    dir_path = Path(obj)
+                    if dir_path.is_dir():
+                        result.extend(sorted(walk_graphql_files(dir_path)))
+                    else:
+                        result.append(dir_path)
+                continue
+            except (ImportError, ModuleNotFoundError):
+                pass
+
+        local_path = Path(source)
+        if local_path.is_dir():
+            result.extend(sorted(walk_graphql_files(local_path)))
+        else:
+            result.append(local_path)
+    return result
+
+
+def get_graphql_schema_from_paths(schema_paths: list[str]) -> GraphQLSchema:
+    """Get graphql schema built from multiple path sources."""
+    resolved = resolve_schema_paths(schema_paths)
+    schema_str = "\n".join(read_graphql_file(p) for p in resolved)
     graphql_ast = parse(schema_str)
     schema: GraphQLSchema = build_ast_schema(graphql_ast, assume_valid=True)
     return schema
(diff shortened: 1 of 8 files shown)


ariadne_codegen/schema.py
```diff
@@ -141,6 +142,53 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
...
+    for source in sources:
+        if (
+            "." in source
+            and "/" not in source
+            and not source.endswith((".graphql", ".graphqls", ".gql"))
+        ):
+            try:
+                module_path, attr = source.rsplit(".", 1)
+                module = importlib.import_module(module_path)
+                obj = getattr(module, attr)
```
DamianCzajkowski: This can raise an AttributeError exception. We need to consider whether we want to catch it and handle it ourselves or leave it as is.

ariadne_codegen/settings.py
```diff
@@ -303,11 +324,16 @@ def used_settings_message(self):
             self._introspection_settings_message() if self.using_remote_schema else ""
         )
 
+        schema_source = self.schema_path or (
```
DamianCzajkowski: This looks like code duplication; you already have a property method for the schema source

DamianCzajkowski (approved):

docs/02-configuration.md
```diff
@@ -18,11 +18,32 @@ queries_path = "queries.graphql"
...
 - `remote_schema_url` - url to graphql server, where introspection query can be perfomed
 
+### `schema_paths` resolution
+
+Each entry is resolved in order:
+
+1. **Dotted Python attribute** (no `/` in the string, not ending with a graphql extension) — the attribute is imported via `importlib`:
+   - **callable** → called, expected to return a list of file paths
+   - **string / `Path`** → treated as a directory and searched recursively for `.graphql`, `.graphqls`, `.gql` files
+2. **Local path** (fallback when import fails, or when the entry contains `/` or has a graphql extension) — a file or directory searched recursively.
```
Kwaidan00: I don't think the docs should contain a description like this, which described the algorithm of calculating imports. 

Instead of that, as a user who tries to configure the tool, I would expect declarative data: 

> for this field, you can define any of the following:
> - absolute import path to the callable which must returns list[str]
> - absolute import path to the variable which must contain the path to the single file
> - absolute import path to the variable which must contain the path to the directory; all files `.graphql`, `.graphqls`, `.gql` from this directory will be included
> - path to the directory;  all files `.graphql`, `.graphqls`, `.gql` from this directory will be included
> - path to the specific file to be used

ariadne_codegen/schema.py
```diff
@@ -141,6 +142,53 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
...
+    Each entry is tried as a dotted Python import path first (e.g.
+    ``pkg.SCHEMA_DIR`` or ``pkg.get_schema_files``). If the import fails the
+    entry is treated as a local filesystem path instead.
+    """
+    result: list[Path] = []
+    for source in sources:
+        if (
+            "." in source
+            and "/" not in source
+            and not source.endswith((".graphql", ".graphqls", ".gql"))
```
Kwaidan00: 1. This list of "graphql extensions" already exists in `walk_graphql_files`. Please do not duplicate the code - if you need to use it here, refactor code to reuse the same list in both places.
2. Did you consider "ask forgiveness not permission" approach? Like "try to parse source as file/dir; if it's not reachable, try to parse it as import path". Your current solution brings some assumptions for the allowed single name. With the `schema_path`, I can configure it to use `my_schema_file.any` and it will work. With your solution - it won't. And you have more checks to do.

ariadne_codegen/schema.py
```diff
@@ -141,6 +142,53 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
...
+                module = importlib.import_module(module_path)
+                obj = getattr(module, attr)
+                if callable(obj):
+                    result.extend(Path(f) for f in obj())
+                elif isinstance(obj, (str, Path)):
+                    dir_path = Path(obj)
+                    if dir_path.is_dir():
+                        result.extend(sorted(walk_graphql_files(dir_path)))
+                    else:
+                        result.append(dir_path)
```
Kwaidan00: This implementation shows you allow not only dir paths in this imported variable, but also a path to a single file. Thus:
1. Docs should mention that (I think I also described it in the comment about docs)
2. This variable name `dir_path` is misleading.

ariadne_codegen/schema.py
```diff
@@ -141,6 +142,53 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
...
+        if local_path.is_dir():
+            result.extend(sorted(walk_graphql_files(local_path)))
+        else:
+            result.append(local_path)
+    return result
+
+
+def get_graphql_schema_from_paths(schema_paths: list[str]) -> GraphQLSchema:
+    """Get graphql schema built from multiple path sources."""
+    resolved = resolve_schema_paths(schema_paths)
```
Kwaidan00, line 224: There is no place you check if the provided values in `resolved` really exists. Please compare it with the current approach with `schema_path`: it runs `assert_path_exists` and raises `InvalidConfiguration(f"Provided path {path} doesn't exist.")` if file does not exist. Your code will crash on `open`.

Please check if every file in list of sources exists and raise `InvalidConfiguration` if not.

ariadne_codegen/schema.py
```diff
@@ -141,6 +142,53 @@ def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
...
+
+        local_path = Path(source)
+        if local_path.is_dir():
+            result.extend(sorted(walk_graphql_files(local_path)))
+        else:
+            result.append(local_path)
+    return result
+
+
+def get_graphql_schema_from_paths(schema_paths: list[str]) -> GraphQLSchema:
```
Kwaidan00, line 222: This code is very similar to the existing `get_graphql_schema_from_path`. Did you consider refactoring code to have only one function with the responsibility "read given files and return `GraphQLSchema` object"?

ariadne_codegen/schema.py
Kwaidan00: The existing code for paring config performs strict validation for settings - which is good.
1. I definitely agree AttributeError should be caught and the `InvalidConfiguration` exception should be raised from it
2. The current `except (ImportError, ModuleNotFoundError): pass` should not silently fail - it should raise `InvalidConfiguration`. If user configured sources list as import from `foo.bar.SCHEMA_DIR_PATH` and this dependency is not reachable (no package, value removed from new version of the package) we definitely should raise an error - because the tool configuration won't produce the expected result.

docs/02-configuration.md
```diff
@@ -18,11 +18,35 @@ queries_path = "queries.graphql"
...
-One of the following 2 parameters is required, in case of providing both of them `schema_path` is prioritized:
+Exactly one of the following 3 parameters is required. They are mutually exclusive - providing more than one raises a configuration error:
 
 - `schema_path` - path to file/directory with graphql schema
+- `schema_paths` - list of schema sources resolved at codegen time; each entry may be a local path (file or directory) or a dotted Python attribute path (`pkg.ATTR` or `pkg.callable`). See details below.
 - `remote_schema_url` - url to graphql server, where introspection query can be perfomed
 
+### `schema_paths` entries
+
+Each entry in `schema_paths` can be any of the following:
```
Kwaidan00: I would say it must, not "can".

docs/02-configuration.md
```diff
@@ -18,11 +18,35 @@ queries_path = "queries.graphql"
...
+
+- **an absolute import path to a callable** that returns a `list[str]` of file paths, eg. `some_pkg.get_schema_files`
+- **an absolute import path to a variable** holding the path to a single schema file, eg. `some_pkg.SCHEMA_FILE`
+- **an absolute import path to a variable** holding the path to a directory - all `.graphql`, `.graphqls` and `.gql` files from it are included, eg. `some_pkg.SCHEMA_DIR`
+- **a path to a directory** - all `.graphql`, `.graphqls` and `.gql` files from it are included, eg. `./schemas/`
+- **a path to a specific file** to be used, eg. `./foo/bar.graphql`
+
+An import path that cannot be resolved (missing package, or the attribute no longer exists) raises a configuration error rather than being silently skipped.
```
Kwaidan00: Please remove it. You declared the lists of possible values, it's rather obvious it will raise an error. Moreover, we do not have similar mentions like "when providing remote_schema_url which is invalid URL, it will raise an error".

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    file1 = tmp_path / "schema.graphql"
+    file1.write_text(schema_str, encoding="utf-8")
+    file2 = tmp_path / "user.graphql"
+    file2.write_text(extra_type_str, encoding="utf-8")
+
+    result = resolve_schema_paths([file1.as_posix(), file2.as_posix()])
+
+    assert file1 in result
+    assert file2 in result
+    assert len(result) == 2
```
Kwaidan00: I don't understand what's the point to have three separate test cases here. The first one checks the single file, the second - a single directory, the last one: two files.

WDYT to have just a single test case, which tests reading local files, and provide as an input a list: single file and single directory.

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    mock_module.SCHEMA_FILE = missing.as_posix()
+    mocker.patch(
+        "ariadne_codegen.schema.importlib.import_module", return_value=mock_module
+    )
+
+    with pytest.raises(InvalidConfiguration):
+        resolve_schema_paths(["some_pkg.SCHEMA_FILE"])
+
+
+def test_resolve_schema_paths_existing_file_skips_importlib(
```
Kwaidan00: Why do you propose to have a test like this? What's the purpose?

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    schema_file = tmp_path / "schema.graphql"
+    schema_file.write_text(schema_str, encoding="utf-8")
+    mock_import = mocker.patch("ariadne_codegen.schema.importlib.import_module")
+
+    resolve_schema_paths([schema_file.as_posix()])
+
+    mock_import.assert_not_called()
+
+
+def test_resolve_schema_paths_accepts_file_with_any_extension(tmp_path, schema_str):
```
Kwaidan00: Again, why this test is needed?

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    schema_file.write_text(schema_str, encoding="utf-8")
+    mock_import = mocker.patch("ariadne_codegen.schema.importlib.import_module")
+
+    result = resolve_schema_paths([schema_file.as_posix()])
+
+    assert result == [schema_file]
+    mock_import.assert_not_called()
+
+
+def test_resolve_schema_paths_accepts_local_dir_with_dots_in_name(
```
Kwaidan00: I understand to show the code works for dots in filenames. But are you sure we absolutely need to have two different test cases for file and for directory?

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    (schemas_dir / "user.graphql").write_text(extra_type_str, encoding="utf-8")
+    mock_import = mocker.patch("ariadne_codegen.schema.importlib.import_module")
+
+    result = resolve_schema_paths([schemas_dir.as_posix()])
+
+    assert {p.name for p in result} == {"schema.graphql", "user.graphql"}
+    mock_import.assert_not_called()
+
+
+def test_resolve_schema_paths_with_path_attribute_pointing_to_file(
```
Kwaidan00: Is this test case just a "file" version of the "directory" test `test_resolve_schema_paths_with_path_attribute_pointing_to_directory`? If so, why it lives here in the file, not directly below the other similar tests?

tests/test_schema.py
```diff
@@ -652,3 +655,252 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+    mocker.patch(
+        "ariadne_codegen.schema.importlib.import_module", return_value=mock_module
+    )
+
+    result = resolve_schema_paths(["some_pkg.SCHEMA_FILE"])
+
+    assert result == [schema_file]
+
+
+def test_resolve_schema_paths_raises_invalid_configuration_for_source_without_dot(
```
Kwaidan00, line 833: I would suggest to keep all the test cases "for this bad data, it raises InvalidConfiguration" in one place. We had some tests like this above, then other test cases, now we have again tests like this.

ariadne_codegen/schema.py
```diff
@@ -133,22 +135,89 @@ def introspect_remote_schema(
...
+
+
+def _build_schema_from_str(schema_str: str) -> GraphQLSchema:
+    """Build a GraphQLSchema from concatenated schema source."""
     graphql_ast = parse(schema_str)
     schema: GraphQLSchema = build_ast_schema(graphql_ast, assume_valid=True)
     return schema
 
 
+def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
```
Kwaidan00, line 150: Did you consider just

```

def get_graphql_schema_from_path(schema_path: str) -> GraphQLSchema:
    return get_graphql_schema_from_paths([Path(schema_path)])

```

With the current solution, bot "from_path" and "from_paths" uses similar blocks, but call them independently.

tests/test_schema.py
```diff
@@ -652,3 +655,210 @@ def test_introspect_remote_schema_uses_default_introspection_settings_when_not_p
...
+):
+    missing = tmp_path / "does_not_exist"  # neither a file/dir nor a dotted path
+
+    with pytest.raises(InvalidConfiguration):
+        resolve_schema_paths([missing.name])
+
+
+def test_resolve_schema_paths_raises_invalid_configuration_for_unexpected_type(mocker):
+    mock_module = mocker.Mock()
+    mock_module.NOT_A_PATH = 123  # not callable, not str/Path
```
Kwaidan00: If you want to validate if the specific members matches the requirements, please note you do not perform so strict validation for the callable: you just trust the conventional protocol defined in docs. `return [Path(f) for f in obj()]` will crash (with no catch and nice `InvalidConfiguration`) for:
- non-iterable returned by `obj()`
- incorrect type passed to `Path`, e.g. if `obj()` returns `[None, 123]`

Kwaidan00 (approved):

DamianCzajkowski (approved):

PR #440, opened 2026-06-29
feat: Allow to use non-httpx HTTP clients in default clients [AR-1]

Kwaidan00:
This PR changes the default clients implementation (both sync and async) to accept not only `httpx` clients as `http_client`.  With this change, user can define its own class or use clients from other libraries (like `requests.Session`) if they fulfill the protocol.

To achieve that:
- the custom `Protocol` was defined and documented
- the default client implementation was adjusted to only use methods and attributes from the defined protocol.


ariadne_codegen/client_generators/dependencies/base_client.py
```diff
@@ -188,22 +209,15 @@ def _execute_json(
         operation_name: Optional[str],
         variables: dict[str, Any],
         **kwargs: Any,
-    ) -> httpx.Response:
-        headers: dict[str, str] = {"Content-type": "application/json"}
```
DamianCzajkowski, line 192: Worth mentioning that we dropped explicit content-type handling, and it becomes an implicit requirement for any client to satisfy the `protocol`

DamianCzajkowski (approved):
Missing changelog entry. It's generated by git-cliff from conventional commits. Please use a proper prefix in the commits and generate the changelog entry with hatch run changelog-update

ariadne_codegen/client_generators/dependencies/base_client.py
Kwaidan00: @DamianCzajkowski I don't think it's needed to be worth to mention in the docs. I checked the commits and at some point we switched from using `httpx.post(json=...)` to `httpx.post(content=...)` to handle serialization on our own; then this overcomplicated code related to headers was added to properly set the content type. Other words, we had to handle the content-type header as the hack for not using `json` parameter.

With the current interface we assume the `http_client` sends `post` requests properly and it's not "client base class" responsibility to handle content-type header, in the same way it's not "client base class" responsibility to handle content-length. Especially we switched to `json=` param.

The only case that could be worth to analyze is the backward compatibility, if someone already uses custom `http_client`. If we aware for this case, we need to consider this PR as breaking change.

Minister944 (approved):

PR #441, opened 2026-07-02
docs: consolidate plugin docs into docs site AR-7

Minister944:
Remove standalone PLUGINS.md and fold its content into
docs/06-plugins/02-custom-plugins.md, expanding the hooks
reference and plugin creating guide.


DamianCzajkowski (approved):

docs/06-plugins/02-custom-plugins.md
```diff
@@ -22,84 +22,239 @@ Every element of the list can enable plugins in 2 ways:
...
+### Sequential pipeline
+
+All plugins run in the order they are listed under `plugins` in `pyproject.toml`. Each plugin receives the **output of the previous plugin** as its input, not the original value. If plugin A and plugin B both implement `generate_client_class`, B sees the `ast.ClassDef` that A already modified.
+
+There is no priority mechanism. Every plugin executes every hook it defines. Order is controlled entirely by the position in the `plugins` list.
+
+### What every plugin instance has
+
+- `self.schema: GraphQLSchema` - the full GraphQL schema, kept up-to-date by `process_schema`
+- `self.config_dict: dict` - the parsed `pyproject.toml` as a dict
```
Kwaidan00: It's already mentioned in the first section on this page. Why do you need to repeat it?

docs/06-plugins/02-custom-plugins.md
```diff
@@ -22,84 +22,239 @@ Every element of the list can enable plugins in 2 ways:
...
+
+### What plugins cannot do
+
+- Control or skip other plugins in the chain
+- Read the output of sibling plugins
+- Prevent the hook chain from continuing
+- Write files directly (transform content through code-string hooks; the framework handles writing)
```
Kwaidan00: I don't think we need to list things plugins "cannot do". I don't see the reasons why should I read that; input/output, described hooks and examples are better way to present what it "can" do.

docs/06-plugins/02-custom-plugins.md
```diff
@@ -22,84 +22,239 @@ Every element of the list can enable plugins in 2 ways:
 
 2. Providing only package name, eg. `my_package`. In this case all plugins from selected package will be used. For class to be accessible this way, it needs to be in package's public api. Plugin has to be importable from package, eg. `from my_package import MyPlugin` has to work.
 
-## Hooks
+## Plugin execution model
 
-### generate_init_module
+### Sequential pipeline
+
+All plugins run in the order they are listed under `plugins` in `pyproject.toml`. Each plugin receives the **output of the previous plugin** as its input, not the original value. If plugin A and plugin B both implement `generate_client_class`, B sees the `ast.ClassDef` that A already modified.
```
Kwaidan00: This `If plugin A and plugin B...` is the same as examples below. Do you need to mention it here as well?

docs/06-plugins/02-custom-plugins.md
```diff
@@ -22,84 +22,239 @@ Every element of the list can enable plugins in 2 ways:
...
-### generate_gql_function
+### process_schema
 
 ```py
-def generate_gql_function(self, function_def: ast.FunctionDef) -> ast.FunctionDef:
+def process_schema(self, schema: GraphQLSchema) -> GraphQLSchema:
 ```
 
-Hook executed on generation of `gql` function.
+Called once at startup, immediately after the schema is loaded from path or URL. `assume_valid` is set to `True` during parsing, so `graphql.assert_valid_schema` runs only after this hook returns. This is the earliest point at which plugins can inspect or modify the schema. After each plugin's `process_schema` returns, `self.schema` is updated on **all** plugin instances.
```
Kwaidan00: Where is this `assume_valid`? I feel like I'm lost with this sentence. If it's part of some bigger mechanism, it would be nice to at least mention how it's related to this hook or where this code lives.

Kwaidan00 (approved):

PR #442, opened 2026-07-07
Add failing test: duplicate typename__ when type implements multiple interfaces

magicmark:
When a concrete type implements multiple interfaces and a query spreads on both with `__typename`, codegen emits the field twice — which makes mypy blow up with `Name "typename__" already defined [no-redef]`.

```graphql
type Foo implements Bar & Baz { id: ID!, name: String! }

query GetFoo {
    foo {
        ... on Bar { __typename id }
        ... on Baz { __typename name }
    }
}
```

Actual:
```python
class GetFooFoo(BaseModel):
    typename__: Literal["Foo"] = Field(alias="__typename")
    id: str
    typename__: Literal["Foo"] = Field(alias="__typename")
    name: str
```

Expected:
```python
class GetFooFoo(BaseModel):
    typename__: Literal["Foo"] = Field(alias="__typename")
    id: str
    name: str
```

Works fine at runtime (pydantic is happy), but mypy won't accept it.

Fix is at https://github.com/magicmark/ariadne-codegen/pull/1 (on my fork for now — can't target a fork branch as base on upstream).

### DIFF
--- a/tests/client_generators/result_types_generator/test_duplicate_typename.py
+++ b/tests/client_generators/result_types_generator/test_duplicate_typename.py
@@ -1,0 +1,69 @@
+import ast
+
+import pytest
+from graphql import OperationDefinitionNode, build_schema, parse
+
+from ariadne_codegen.client_generators.constants import TYPENAME_ALIAS
+from ariadne_codegen.client_generators.result_types import ResultTypesGenerator
+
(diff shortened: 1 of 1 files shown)


magicmark:
@mirumee any thoughts?

PR #444, opened 2026-07-10
feat: Allow to use custom client for fetching remote schema [AR-2]

Kwaidan00:
This PR allows to define `remote_schema_http_client_path` to point the module, class or object which will be used as HTTP client to fetch the schema introspection from remote server.


DamianCzajkowski (approved):
LGTM

PR #446, opened 2026-07-13
feat: refactor documentation AR-8

Minister944:
Restructures the docs from a flat, README-derived layout into a categorized Docusaurus site, and slims README.md into a landing page. Content was split out of the README, then expanded and corrected against the source code - several settings and behaviours were missing or stale.

New docs/ structure:

- docs/
  - 01-introduction.md
  - 02-guides/     step-by-step, schema sources, using the client, subscriptions, file uploads, custom scalars, extending types, multiple clients, OpenTelemetry, async vs sync, schema generation, custom operations
  - 03-reference/  configuration, generated code dependencies
  - 04-plugins/    (renamed from 06-plugins; contents unchanged)
  - 05-community/  contributing, versioning policy


docs/01-introduction.md
Minister944, line 1: I wasn't able to set this int reference for the readme, so here's a copy-paste

EXAMPLE.md
Minister944, line 1: Do we need this file?

docs/02-guides/02-schema-sources.md
```diff
@@ -0,0 +1,96 @@
...
+from a local file/directory, from installed Python packages, or by introspecting a
+remote GraphQL server.
+
+Exactly one of the following settings is required - `schema_path`, `schema_paths`
+and `remote_schema_url` are **mutually exclusive**, so only one schema source may be
+used at a time:
+
+- `schema_path` - path to file/directory with GraphQL schema
+- `schema_paths` - list of local paths and/or installed-package sources to build the schema from
+- `remote_schema_url` - url to GraphQL server, where introspection query can be performed
```
Kwaidan00: url -> URL

docs/02-guides/02-schema-sources.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Local schema file
```
Kwaidan00, line 19: I think this section should mention the behavior for directory: `ariadne-codegen` will read all the files with the specific extension.

docs/02-guides/02-schema-sources.md
```diff
@@ -0,0 +1,96 @@
...
+
+```toml
+[tool.ariadne-codegen]
+schema_path = "schema.graphql"
+queries_path = "queries.graphql"
+```
+
+## Loading schema from installed packages
+
+`schema_paths` lets you pull type definitions from installed Python packages
```
Kwaidan00: I would say the most important use case for `schema_paths` is to point to multiple schema files in the different locations. AND it also allows to import from Python packages.

docs/02-guides/02-schema-sources.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+The following options control how the introspection request is performed:
+
+- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$AUTH_TOKEN"}`
+- `remote_schema_verify_ssl` (defaults to `true`) - a flag that specifies whether to verify ssl while introspecting remote schema
+- `remote_schema_timeout` (defaults to `5`) - timeout in seconds while introspecting remote schema
+- `remote_schema_http_client_path` (defaults to `None`) - dotted import path to a custom HTTP client used only for the introspection request, instead of the default `httpx` client
+
+## Introspection query options
```
Kwaidan00: It's the additional section that describes the `Remote schema introspection` strategy. Logically it should be below it:
```suggestion
### Introspection query options
```

docs/02-guides/03-using-generated-client.md
```diff
@@ -0,0 +1,119 @@
...
+from graphql_client.client import Client
+```
+
+### Passing headers to client
+
+Client (with default base client), takes passed headers and attaches them to every sent request.
+
+```py
+client = Client("https://example.com/graphql", {"Authorization": "Bearer token"})
+```
```
Kwaidan00, line 29: You showed how to import it. Here's the more advanced usage (with headers). Later in this doc is a context manager. It lacks the basic case of using Client (I know, it could be silly, but I think it should be there for completeness).

So after the example import, I propose something like:
> You can initialize the imported client with the server URL: client = Client("https://example.com/graphql

Then, without the `### Passing headers to client` header, the current paragraph about adding headers.

docs/02-guides/03-using-generated-client.md
```diff
@@ -0,0 +1,119 @@
...
+
+For more complex scenarios, you can pass your own http client:
+
+```py
+client = Client(http_client=CustomComplexHttpClient())
+```
+
+`CustomComplexHttpClient` needs to be an instance of `httpx.AsyncClient` for async client, or `httpx.Client` for sync.
```
Kwaidan00: First of all, it's no longer true: after the recent changes it does not require to be `httpx` class instances. Secondly, I think it's described below, so you can just keep here
> For more complex scenarios, see [link to section below]

docs/02-guides/03-using-generated-client.md
```diff
@@ -0,0 +1,119 @@
...
+
+The `ws_*` arguments only exist on the async base client (subscriptions are not
+supported by the synchronous client). See [Subscriptions](./04-subscriptions.md) for
+details. The synchronous base client (`async_client = false`) accepts only `url`,
+`headers` and `http_client` - see [Async vs sync client](./10-async-vs-sync.md).
+
+If you enable the OpenTelemetry base client (`opentelemetry_client = true`), the
+constructor also accepts tracing arguments - see [Open Telemetry](./09-opentelemetry.md).
+
+## Using the client as a context manager
```
Kwaidan00: This section should live before `Client configuration` section.

With the current layout, we have usage description, then configuration, then this section about usage, then advanced configuration again.

docs/02-guides/03-using-generated-client.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+and `with` for the synchronous client:
+
+```py
+with Client(url="https://example.com/graphql") as client:
+    result = client.list_all_users()
+```
+
+## Advanced HTTP configuration
```
Kwaidan00: This doc completely forgot about the recent changes with custom clients.

If you want to keep this section about configuring HTTPX clients, I'm ok with that, but please rename it to sth like "Advanced HTTP configuration with httpx clients". And please add the next section `Advanced HTTP configuration with custom client`, with the content similar to https://github.com/mirumee/ariadne-codegen/blob/ddc53384d37b32ce8bc91af8742aed02d5600b5d/README.md?plain=1#L249-L303

docs/02-guides/03-using-generated-client.md
```diff
@@ -0,0 +1,119 @@
...
+    ws_url="wss://example.com/graphql",
+    ws_headers={"Authorization": "Bearer token"},
+    ws_origin="https://example.com",
+    ws_connection_init_payload={"Authorization": "Bearer token"},
+)
+```
+
+- `url` (defaults to `""`) - HTTP endpoint the client sends operations to.
+- `headers` (defaults to `None`) - headers attached to every HTTP request. When no `http_client` is supplied, they are also used to build the default `httpx.AsyncClient`.
+- `http_client` (defaults to `None`) - your own HTTP client instance. Must be an `httpx.AsyncClient` for the async client, or `httpx.Client` for the sync client. When provided, `url`/`headers` handling for the transport is up to you.
```
Kwaidan00: If you want to mention `http_client` parameter here, fix this line; it's no longer true. I suggest to keep it short:
> your own HTTP client instance. See [link to advanced with HTTPX client] and [link to advanced with custom client] for details.

docs/02-guides/04-subscriptions.md
```diff
@@ -0,0 +1,102 @@
...
+
+Each server push yields a new `GetUsersCounter` instance; the loop runs until the
+server closes the subscription (or you `break` out of it).
+
+## Connection arguments
+
+Arguments `ws_origin` and `ws_headers` are added as headers to the handshake
+request, and `ws_connection_init_payload` is used as the payload of the
+[ConnectionInit](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit)
+message. See [Using generated client](./03-using-generated-client.md#client-configuration)
```
Kwaidan00: The `ws_url` is mentioned earlier, and then you mentioned `ws_origin`, `ws_headers`, `ws_connection_init_payload`. There is no more arguments described in the linked page. So this link can be dropped.

docs/02-guides/07-extending-types.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+You don't need to declare the directive in your schema - `ariadne-codegen` registers
+it automatically before processing your operations, so you can use `@mixin` right
+away.
+
+Both arguments must be string literals; if either `from` or `import` is missing (or
+a non-string value is passed), generation fails with a `ParsingError`.
+
+This directive can be used along with the `files_to_include` option to extend the
```
Kwaidan00: Can be used or must be used? As I understand, the module `from` must be reachable, so if I use the `from ".mixins"`, I have to include it in the `files_to_include`. Am I right?

docs/02-guides/09-opentelemetry.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Open Telemetry
+---
+
+# Open Telemetry
+
+When the config option `opentelemetry_client` is set to `true` then the default,
+included base client is replaced with one that implements opt-in Open Telemetry
+support. By default this support does nothing, but when the `opentelemetry-api`
```
Kwaidan00: I suggest to add a PyPI link to this package.

docs/02-guides/10-async-vs-sync.md
```diff
@@ -0,0 +1,94 @@
...
+`async` method and the client inherits from `AsyncBaseClient`.
+
+To generate a synchronous client instead, set `async_client` to `false`:
+
+```toml
+[tool.ariadne-codegen]
+async_client = false
+```
+
+- `async_client` (defaults to `true`) - default generated client is `async`, change this option to `false` to generate a synchronous client instead
```
Kwaidan00: This line seems to be not needed here.

docs/02-guides/10-async-vs-sync.md
```diff
@@ -0,0 +1,94 @@
...
+
+## What changes
+
+|                        | Async (default)                     | Sync (`async_client = false`)     |
+| ---------------------- | ----------------------------------- | --------------------------------- |
+| Base class             | `AsyncBaseClient`                   | `BaseClient`                      |
+| Operation methods      | `async def` (must be `await`ed)     | plain `def`                       |
+| Context manager        | `async with`                        | `with`                            |
+| Custom `http_client`   | `httpx.AsyncClient`                 | `httpx.Client`                    |
```
Kwaidan00: It's not true. For async/sync the custom client should just implement other protocols.

docs/02-guides/10-async-vs-sync.md
```diff
@@ -0,0 +1,94 @@
...
+asyncio.run(main())
+```
+
+```py
+# sync
+from graphql_client.client import Client
+
+with Client(url="https://example.com/graphql") as client:
+    users = client.list_all_users()
+```
```
Kwaidan00: Please remove - it's the repetition of 03-using-generated-client.

docs/02-guides/11-schema-generation.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+ariadne-codegen graphqlschema
+```
+
+`graphqlschema` mode reads configuration from the same place as the client, but uses
+only the `schema_path`, `schema_paths`, `remote_schema_url`, `remote_schema_headers`,
+`remote_schema_verify_ssl`, `remote_schema_timeout`, `remote_schema_http_client_path`,
+and `introspection_*` options to retrieve the schema (see
+[Schema sources](./02-schema-sources.md)) and the `plugins` option to load plugins.
```
Kwaidan00: I would consider rephrasing this part with bullet list - it will be more readable.

docs/02-guides/11-schema-generation.md
```diff
@@ -0,0 +1,74 @@
...
+
+Before writing the file, the loaded schema is passed through each plugin's
+`process_schema` hook and then validated, so this mode can also be used to apply
+plugin transformations to a schema or to fail early on an invalid one.
```
Kwaidan00: I suggest to add something like "see [link to plugins] for details".

docs/02-guides/12-custom-operation-builder.md
```diff
@@ -6,6 +6,28 @@ title: Custom operation builder
 
 The custom operation builder allows you to create complex GraphQL queries in a structured and intuitive way.
 
+## Enabling
```
Kwaidan00: I don't think you need this header. Instead of that, you can just
> The custom operation builder allows you to create complex GraphQL queries in a structured and intuitive way. This feature is disabled by default. To enable it, set `enable_custom_operations = true` in your config. (...)

docs/02-guides/12-custom-operation-builder.md
```diff
@@ -73,6 +95,32 @@ asyncio.run(get_products())
    1. The client.query(...) method is called with the built queries and an operation name "get_products".
    2. This method sends the queries to the server and retrieves the response.
 
+Unlike the per-operation client methods, `client.query(...)` and
+`client.mutation(...)` return the raw response as a `dict[str, Any]` - the result is
+not parsed into a generated Pydantic model.
+
+## Building mutations
```
Kwaidan00: The section above is `## Example Code` and shows the custom query.

I think the section above could be renamed to e.g. `Example: custom queries`, and this part could be named `Example: custom mutations`

docs/03-reference/01-configuration.md
```diff
@@ -18,45 +18,25 @@ queries_path = "queries.graphql"
...
-- `remote_schema_verify_ssl` (defaults to `true`) - a flag that specifies wheter to verify ssl while introspecting remote schema
+- `remote_schema_headers` - extra headers that are passed along with introspection query, eg. `{"Authorization" = "Bearer token"}`. To include an environment variable in a header value, prefix the variable with `$`, eg. `{"Authorization" = "$AUTH_TOKEN"}`
+- `remote_schema_verify_ssl` (defaults to `true`) - a flag that specifies whether to verify ssl while introspecting remote schema
 - `remote_schema_timeout` (defaults to `5`) - timeout in seconds while introspecting remote schema
-- `remote_schema_http_client_path` - absolute import path to the HTTP client class used to introspect remote schema. If not provided, default `httpx` client is used.
+- `remote_schema_http_client_path` (defaults to `None`) - dotted import path to a custom HTTP client (a class/callable that returns one, or a ready client object) providing an `httpx`-compatible `post` interface, used only for the remote schema introspection request. If unset, the default `httpx` client is used.
```
Kwaidan00: This new description is not correct. Please bring back the previous description.

README.md
```diff
@@ -4,555 +4,96 @@
...
-- `hatch test` – tests with coverage (default Python 3.10)
-- `hatch test -a -p` – tests across all supported Python versions
-- `hatch run check` – full check (format, typecheck, tests)
 
 ## Contributing
 
-We welcome all contributions to Ariadne! If you've found a bug or issue, feel free to use [GitHub issues](https://github.com/mirumee/ariadne-codegen/issues). If you have any questions or feedback, don't hesitate to catch us on [GitHub discussions](https://github.com/mirumee/ariadne/discussions/).
-
-For guidance and instructions, please see [CONTRIBUTING.md](CONTRIBUTING.md).
+Contributions are welcome! See [Contributing](docs/05-community/01-contributing.md) for how to report bugs, work on issues, and open pull requests.
```
Kwaidan00: But this page points to CONTRIBUTING.md. For README on GitHub, shouldn't it point to CONTRIBUTING.md directly?

Kwaidan00 (approved):

DamianCzajkowski (approved):

EXAMPLE.md
DamianCzajkowski: leave it for now

PR #447, opened 2026-07-20
feat: defer model build

DamianCzajkowski:
Adds an opt-in `defer_model_build` setting (default `false`) that makes importing
a generated package much faster for large schemas, plus the build/tooling and
correctness fixes the change that rests on.

### What it adds

**`defer_model_build` setting** when enabled:
- sets `defer_build=True` on the generated `BaseModel`, so each model's core
  The schema is built on first use instead of at import time;
- drops the eager `Model.model_rebuild()` calls from the generated modules.

For a large schema, this cuts import time roughly in half, since importing the
package no longer builds every Pydantic model up front. It's fully opt-in and
changes no generated code until you set it. Requires `pydantic >= 2.9` in the
environment that runs the generated package (`defer_build` is not a `ConfigDict`
key before 2.5).

```toml
[tool.ariadne-codegen]
defer_model_build = true

### CODE REVIEW

File: ariadne_codegen/client_generators/package.py [hunk 1]
```diff
@@ -193,9 +212,56 @@ def generate(self) -> list[str]:
...
+            group = [
+                pending
+                for pending in self._pending_files
+                if pending.remove_unused_imports is remove_unused_imports
+            ]
+            formatted = format_many(
+                [pending.code for pending in group],
+                remove_unused_imports=remove_unused_imports,
+            )
+            for pending, code in zip(group, formatted, strict=True):
```
Kwaidan00: I think `pending_file` would be more readable than `pending`.

tests/performance/test_codegen_performance.py
```diff
@@ -0,0 +1,157 @@
...
+    assert large_files > small_files * 2, "the two packages must differ in size"
+    assert large_spawns == small_spawns, (
+        f"ruff spawn count scales with the package: "
+        f"{small_files} files -> {small_spawns} spawns, "
+        f"{large_files} files -> {large_spawns} spawns"
+    )
+    assert large_spawns <= MAX_RUFF_SPAWNS
+
+
+def test_ruff_is_not_spawned_through_a_python_interpreter(tmp_path):
```
Kwaidan00: Do you really need this large test case just to check the return value of `_ruff_command` util?
Moreover, the util suggests (by defining fallback operation) it's fine to use ruff with Python interpreter; with the assertion in this test, it makes the tests failing for the fallback value.

tests/performance/test_codegen_performance.py
```diff
@@ -0,0 +1,157 @@
...
+
+    calls, _ = _generate_recording_spawns(tmp_path)
+
+    assert _ruff_spawns(calls), "expected generation to run ruff"
+    for call in calls:
+        argv = call[0][0]
+        assert "-m" not in argv[:2], f"ruff spawned through an interpreter: {argv[:3]}"
+
+
+def test_batched_formatting_beats_formatting_each_file(request, monkeypatch, tmp_path):
```
Kwaidan00: Do you really need this test case? It seems to be a test case that checks which ruff call is better. It sounds like something to check once before the implementation, not a test case included in the test suite.

DamianCzajkowski: Good catch

ariadne_codegen/client_generators/package.py
DamianCzajkowski: Agreed

tests/performance/test_codegen_performance.py
DamianCzajkowski: Agreed

Kwaidan00 (approved):

Minister944 (approved):

PR #449, opened 2026-07-21
feat: alias generator

DamianCzajkowski:
## What

Opt-in `use_alias_generator`: sets `alias_generator=to_camel` on the generated
`BaseModel` and drops the explicit `Field(alias=...)` from every field whose
GraphQL name `to_camel` can reconstruct. Those calls each build two `FieldInfo`
objects at import, so removing them speeds up the import of large generated packages.

Nothing changes on the wire; names `to_camel` can't reconstruct keep an explicit
alias: `__typename`, keyword-escaped names (`list_`), acronyms (`productID`), and
already-snake_case schemas (`some_field`).

## Notes

- Off by default; output is byte-identical until you set it.
- Requires **pydantic >= 2.8** in the generated package (`to_camel` mangled
  camelCase before then); the build floor is bumped.
- Saves nothing with `convert_to_snake_case = false`; the CLI now says so.

## Also includes: a fragment forward-reference fix

This branch also carries `fix: keep fragment forward references resolvable in
operation modules`

An operation that spreads a fragment generates `class OpNode(Fragment): pass`,
but Pydantic resolves the forward references it *inherits* against `OpNode`'s own
module, where the fragment's nested classes aren't imported. This raised
`PydanticUndefinedAnnotation` / `is not fully defined` on first use. It bit on
Python 3.10 already, and **with `defer_model_build` on every supported version**
(deferring the build removes the eager `model_rebuild()` that used to mask it).

The fix emits the inherited names into the operation module with `# noqa: F401`.


docs/08-fragments-and-forward-references.md
```diff
@@ -0,0 +1,22 @@
...
+    ProductListItemThumbnail,  # noqa: F401
+)
+
+class GetProductsProductsEdgesNode(ProductListItem):
+    pass
+```
+
+Pydantic resolves the annotations a subclass inherits against the subclass's own module, so any class the fragment refers to by forward reference has to be importable there. That is what the second import is for. It carries `# noqa: F401` because the name only appears inside an inherited annotation, and your linter would otherwise call it unused.
+
+This is emitted for every generated client, not only with `defer_model_build`.
```
Kwaidan00: This sentence sounds like a part of the page about defer_model_build. However, this is the separate page that covers another topic. Please remove it or rephrase it (e.g. "this is emitted for every generated client, regardless of the settings")

tests/test_utils.py
```diff
@@ -279,105 +295,172 @@ def test_process_name_returns_name_returned_from_plugin_for_name_with_only_under
...
+        ({"alias_generator": True}, "alias_generator=to_camel"),
+    ],
 )
-def test_rewriting_base_model_keeps_the_generated_file_header(rewrite):
-    """The rewrites round-trip through `ast.unparse`, which drops comments."""
-    code = dedent(
-        """\
-        # Generated by ariadne-codegen
-        # Source: queries.graphql
+def test_rewrite_base_model_adds_a_missing_config_kwarg(settings, expected):
```
Kwaidan00: Naming. I wouldn't say it's missing - you test the function that adds requested things.

tests/test_utils.py
```diff
@@ -279,105 +295,172 @@ def test_process_name_returns_name_returned_from_plugin_for_name_with_only_under
...
     assert result.startswith(
         "# Generated by ariadne-codegen\n# Source: queries.graphql"
     )
+
+
+def test_rewrite_base_model_without_settings_returns_the_source_untouched():
+    assert rewrite_base_model(BASE_MODEL_SOURCE) is BASE_MODEL_SOURCE
+
+
+def test_rewrite_base_model_applies_every_setting_in_one_pass():
```
Kwaidan00: Can't it be just an another parametrized case for `test_rewrite_base_model_adds_a_missing_config_kwarg`?

tests/test_utils.py
```diff
@@ -279,105 +295,172 @@ def test_process_name_returns_name_returned_from_plugin_for_name_with_only_under
...
+    )
+
+    result = rewrite_base_model(code, defer_build=True)
 
     assert result.startswith(
         "# Generated by ariadne-codegen\n# Source: queries.graphql"
     )
+
+
+def test_rewrite_base_model_without_settings_returns_the_source_untouched():
```
Kwaidan00, line 388: Again, can't it be just an another parametrized case for `test_rewrite_base_model_adds_a_missing_config_kwarg`?

tests/test_utils.py
```diff
@@ -279,105 +295,172 @@ def test_process_name_returns_name_returned_from_plugin_for_name_with_only_under
...
+        BASE_MODEL_SOURCE, forbid_extra=True, defer_build=True, alias_generator=True
+    )
+
+    assert "from pydantic.alias_generators import to_camel" in result
+    for kwarg in ('extra="forbid"', "defer_build=True", "alias_generator=to_camel"):
+        assert kwarg in result, kwarg
+    assert "populate_by_name=True" in result
+
+
+def test_rewrite_base_model_only_imports_to_camel_when_it_is_used():
```
Kwaidan00: I think a single parametrized test that applies different settings would be sufficient to show that the tested function adds a line for given argument.

DamianCzajkowski: Agreed also on this one

tests/test_utils.py
DamianCzajkowski: Agreed on this one

tests/test_utils.py
DamianCzajkowski: Here, I would rather keep it separate

DamianCzajkowski: why kept separate? Because it asserts object identity, not substring presence

Kwaidan00 (approved):

Minister944 (approved):

PR #450, opened 2026-07-21
feat: lazy imports

DamianCzajkowski:
Adds an opt-in `lazy_imports` setting. The generated `__init__.py` imports a
module the first time a name from it is used (PEP 562 `__getattr__`) instead of
importing every module up front. This skips defining the models an application
never touches — the biggest win when you use a small slice of a large schema.

It turns on `ClientForwardRefsPlugin` for you, because it needs both halves: the
lazy `__init__` stops the package importing every module, and the plugin stops
`client.py` importing the input types it only names in annotations. Either alone
still pulls the models in. The plugin is appended after any plugin you configured,
since it rewrites the client module they produce.

## Breaking

`refactor!: bind fragment-only result classes straight to the fragment` changes
the shape of generated result classes: an operation that only spreads one fragment
is emitted as `OpNode = Fragment` instead of `class OpNode(Fragment): pass`.
Regenerate the client after upgrading; runtime behavior is unchanged.

## Notes

- Off by default; output is byte-identical until you set it.
- Deferring an import defers what it would have raised; a broken custom mixin now
  surfaces on first use of the name, not at `import your_package`. `__all__` is
  unchanged, so `from pkg import *` and explicit imports behave as before.
- Composes with `defer_model_build` and `use_alias_generator`; all three together
  give the largest speed-up. See docs/07-improving-import-performance.md.


ariadne_codegen/client_generators/init_file.py
```diff
@@ -21,26 +52,83 @@ def add_import(self, names: list[str], from_: str, level: int = 0) -> None:
...
+        return sorted(names)
+
+    def _generate_all_assign(self, lineno: int) -> ast.Assign:
+        return ast.Assign(
+            targets=[ast.Name(id=ALL_NAME)],
+            value=ast.List(
+                elts=[ast.Constant(value=n) for n in self._get_exported_names()]
+            ),
+            lineno=lineno,
+        )
```
Kwaidan00: It would be more readable to define functions below the first usage; for this case, it should be:
- `_generate_eager_module`
- `_generate_all_assign`
- `_get_exported_names`

docs/07-improving-import-performance.md
```diff
@@ -4,30 +4,40 @@ title: Improving import performance
...
+
+`lazy_imports` turns the `ClientForwardRefsPlugin` on for you, and it needs to: the lazy `__init__` stops the package importing every module, and the plugin stops `client.py` importing the input types it only names in annotations. With either one alone the other still pulls the models in, and the import costs what it always did. The plugin is appended after any plugin you configured yourself, because it rewrites the client module they produce; listing it yourself as well changes nothing.
+
+Deferring an import defers what it would have raised. A module that fails to import - a broken [custom mixin](https://github.com/mirumee/ariadne-codegen/blob/main/README.md#extending-models-with-custom-mixins), say - surfaces when the name is first used rather than at `import your_package`. `__all__` is unchanged, so `from your_package import *` and any explicit import of a name still behave as before.
```
Kwaidan00: Please remember to update the link when PR with docs will be merged.

ariadne_codegen/client_generators/init_file.py
DamianCzajkowski: Done

docs/07-improving-import-performance.md
DamianCzajkowski: Thanks! and done

Minister944 (approved):

Kwaidan00 (approved):

PR #457, opened 2026-07-29
feat: Models only mode [AR-4]

Kwaidan00:
This PR adds a new mode to `ariadne-codegen`: to generate only types, with no boilerplate code (client, BaseClient, exceptions).

This change closes the following PRs with the same feature, but different implementation approaches:
- https://github.com/mirumee/ariadne-codegen/pull/436
- https://github.com/mirumee/ariadne-codegen/pull/419

Closes https://github.com/mirumee/ariadne-codegen/issues/418.

## Acknowledgements

This ticket improves the code from #436, with the same goal, test cases and usage, but with different implementation details. Thanks @amin-farjadi for your contribution! 🙏

PR #459, opened 2026-08-07
fix: use include_all_inputs in models_only mode

Kwaidan00:
This PR fixes the behavior of `models_only` mode, introduced in https://github.com/mirumee/ariadne-codegen/pull/457. It allows to set `include_all_inputs` in this mode and generate used inputs only.


DamianCzajkowski (approved):
LGTM

